PR #356, opened 2016-08-29
Hotfix/include favicons in bundle

lraivio:
Includes favicon icons in the main bundle.
Uses better favicon for Espoo.


YoYuUm:
- [ ] There is a typo in the commit name.

PR #357, opened 2016-08-30
Feature/338 update npm packages - part 1

### LINKED ISSUE #339: Update to Babel version 6

lraivio:
This PR:
- Updates most of the project npm packages
- Updates Babel to version 6.x
- Makes the code conform mostly to the updated airbnb JavaScript styleguide
- Improves test configuration and coverage

Refs #338.
Closes #339.


app/actions/__tests__/searchActions.spec.js
```diff
@@ -1,8 +1,8 @@
 import { expect } from 'chai';
 import simple from 'simple-mock';
 
-import searchActions from 'actions/searchActions';
-import apiUtils from 'utils/APIUtils';
+import { searchResources } from 'actions/searchActions';
+import * as apiUtils from 'utils/APIUtils';
```
YoYuUm: Why in tests we have to import modules using `* as`? 

I guess this line ```import { searchResources } from 'actions/searchActions';``` is ok in the tests but it wouldn't be ok out of them?

lraivio: Actions and utils are exported using named exports so they can be imported with 
```javascript
import { searchResources } from 'actions/searchActions';
```
both in tests and in regular files. 

The `import * as apiUtils from 'utils/APIUtils';` is used because otherwise there is no way to mock a single method from the utils.

Exporting only the whole utils or actions module might be better but that would have required quite extensive changes in every file that uses any of the actions or util methods. That's why I didn't do it yet.

YoYuUm: Oh, I guess the wildcard is needed because the utils doesn't have an export default. What's the reason for that?

PR #386, opened 2016-09-13
Feature/385 search page date header

### LINKED ISSUE #385: Search page - add date header

lraivio:
Closes #385.

### DIFF
--- a/app/screens/shared/date-header/DateHeaderComponent.js
+++ b/app/screens/shared/date-header/DateHeaderComponent.js
@@ -21,29 +21,20 @@ class DateHeaderComponent extends Component {
     const { date, onChange } = this.props;
     const dateString = moment(date).format('dddd, LL');
 
-    const iconButtonStyles = {
-      border: 'none',
-      backgroundColor: 'transparent',
-    };
-
     return (
-      <h3
-        className="date-header"
-        id="date-header"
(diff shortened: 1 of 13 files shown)


app/screens/shared/date-header/DateHeaderComponent.js
```diff
@@ -32,18 +32,22 @@ class DateHeaderComponent extends Component {
...
-        </button>
-        <button
-          onClick={() => onChange(addToDate(date, 1))}
-          style={Object.assign({}, iconButtonStyles, { float: 'right' })}
-        >
-          <Glyphicon glyph="chevron-right" />
-        </button>
+        {onChange && (
+          <button
+            onClick={() => onChange(addToDate(date, -1))}
```
YoYuUm: I know this is not related to the changes on this PR, but would say it's not so intuitive to use an onChange this way. I would rather have two onChange functions in the prop, and pass the prop like onClick={buttonNameOnClick}

Maybe it would be clear just changing onChange name for something a more descriptive.

lraivio: Very good point!
I think I could add another commit to this PR and fix that.

lraivio:
@YoYuUm: I made the required changes.

Issue #368, opened 2016-09-05
Search page - update search results styles and info (0.39)


lraivio:
See mockups page 9.

PR #391, opened 2016-09-14
Feature/368 improve search results - part 1

lraivio:
This PR:
- adds a general ResourceList component than is used in the search page for displaying search results. It can also later be used in the home page to display resources near the user.
- Refactors search results so it's easier to implement changes in #368.
- Does not change the appearance of the current search results.

Refs #368.


YoYuUm:
I think the PR is good. Probably the change I just commented  could be done in the next PR.

app/screens/shared/resource-list/ResourceListItem.spec.js
```diff
@@ -4,16 +4,16 @@ import sd from 'skin-deep';
```
lraivio: I just copied the old tests for this one. But your right, it's a lot nicer to have the tests done with enzyme so I made the change already in this PR.

PR #422, opened 2016-09-28
Move icons to ResourceIcons and add price (refs #415)

### LINKED ISSUE #415: Show resource price in resource list items
The price can be a range for example 5 - 50e/h or it can just be one value for example 10e/h.

Resource will have `min_price_per_hour` and `max_price_per_hour` containing the price in strings or a null if price is not set.

See mockups page 9.

YoYuUm:
Closes #415

### DIFF
--- a/app/screens/shared/resource-list/ResourceIcons.js
+++ b/app/screens/shared/resource-list/ResourceIcons.js
@@ -20,7 +20,7 @@ function getHourlyPrice({ minPricePerHour, maxPricePerHour }) {
   if (!(minPricePerHour || maxPricePerHour)) {
     return null;
   }
-  if (minPricePerHour && maxPricePerHour) {
+  if ((minPricePerHour && maxPricePerHour) && (minPricePerHour !== maxPricePerHour)) {
     return `${Number(minPricePerHour)} - ${Number(maxPricePerHour)}e/h`;
   }
   const priceString = maxPricePerHour || minPricePerHour;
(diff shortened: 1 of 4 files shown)


app/screens/shared/resource-list/ResourceIcons.spec.js
```diff
@@ -0,0 +1,135 @@
...
+    it('renders correct text', () => {
+      expect(spanText.text()).to.equal('2h');
+    });
+
+    it('is not rendered if prop is not passed', () => {
+      expect(wrapperNoProps.find({ glyph: 'time' })).to.have.length(0);
+    });
+  });
+
+  describe('time icon', () => {
```
lraivio: - [ ] This should be "euro icon" or maybe "price"

app/screens/shared/resource-list/ResourceIcons.js
```diff
@@ -0,0 +1,49 @@
...
+    </span>
+  );
+}
+
+function getHourlyPrice({ minPricePerHour, maxPricePerHour }) {
+  if (!(minPricePerHour || maxPricePerHour)) {
+    return null;
+  }
+  if (minPricePerHour && maxPricePerHour) {
+    return `${Number(minPricePerHour)} - ${Number(maxPricePerHour)}e/h`;
```
lraivio: - [ ] You could also add a check for a case where `minPricePerHour` and `maxPricePerHour` are equal for example both '10.00' so we avoid displaying '10 - 10e/h'. In those cases only show '10e/h'

app/screens/shared/resource-list/ResourceIcons.spec.js
YoYuUm: I ammended the commit with this change.

lraivio:
Nice! Looks good 👍

Issue #421, opened 2016-09-27
Admin resource page shows wrong availability time before resource opens


lraivio:
Show text "Suljettu" if the resource has not yet opened.

YoYuUm:
- [ ] Show suljetu if the resource is not yet open

lraivio:
Closed in #423.

PR #423, opened 2016-09-28
Show "Closed" if resource is not yet opened

YoYuUm:
Refs #421

### DIFF
--- a/app/components/resource/ResourcesTableItem.js
+++ b/app/components/resource/ResourcesTableItem.js
@@ -35,8 +35,9 @@ class ResourcesTableItem extends Component {
     } = this.props;
     const { closes, opens } = getOpeningHours(resource);
     const now = moment();
+    const closedToday = (closes && opens) === null;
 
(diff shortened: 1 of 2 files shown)


app/components/resource/ResourcesTableItem.js
```diff
@@ -33,17 +33,19 @@ class ResourcesTableItem extends Component {
       nextReservation,
       resource,
     } = this.props;
+    const { closes, opens } = getOpeningHours(resource);
+    const now = moment();
+
+    if ((now < moment(opens)) || (moment(closes) < now)) {
```
lraivio: - [ ] If resources are closed for the whole day their openinghours are `{ closes: null, opens: null }`, for example the [3d printer](http://localhost:3000/resources/auungfvj7qba/reservation).

So you'll have to add a check for that as well. Something like:
```javascript
if (!closes || !opens) {
  return <td className="resource-table-row availability reserved">Suljettu</td>;
}
```

PR #432, opened 2016-10-05
Feature/430 add space before units

### LINKED ISSUE #430: Add space between value and unit in prices and times
- [ ] Price should be written as x €/h and hours should be written as x h
- [ ] Change `e/h` in prices to `€/h`

YoYuUm:
closes #430


lraivio:
Looks good 👍

PR #500, opened 2017-02-15
Feature/494 login inside popup

### LINKED ISSUE #494: Put login inside a popup

lraivio:
- Improves node server side code.
- Moves login to a popup avoid any unnecessary redirects and page reloads keeping the user on the page where they currently are.

Closes #494.


app/actions/authActions.js
```diff
@@ -0,0 +1,45 @@
...
+function fetchAuthState() {
+  const params = { t: +new Date() };
+  return {
+    [CALL_API]: {
+      types: [
+        types.API.AUTH_GET_REQUEST,
+        types.API.AUTH_GET_SUCCESS,
+        types.API.AUTH_GET_ERROR,
+      ],
+      endpoint: `/auth?t=${params.t}`,
```
FelixLoether: - [ ] Does the t param just exist to prevent caching? If so, wouldn't it be better to handle that in Express by adding headers that prevent caching?

server/auth/index.js
```diff
@@ -30,8 +30,14 @@ router.get('/auth', (req, res) => {
...
 
 router.get('/login/helsinki/return',
-  passport.authenticate('helsinki', { failureRedirect: '/login' }),
-  (req, res) => res.redirect('/')
+  passport.authenticate(
+    'helsinki',
+    { failureRedirect: '/login' }
+  ),
+  (req, res) => {
+    const js = 'setTimeout(function() {if(window.opener) { window.close(); } else { location.href = "/"; } }, 300);';
```
FelixLoether: - [ ] Why is a timeout of 300 ms required?

app/actions/authActions.js
lraivio: Good point. Added required headers on Express.

server/auth/index.js
lraivio: Yeah, not needed. Good catch.

Issue #499, opened 2017-02-13
Making reservations from admin resources page


FelixLoether:
Steps to make a reservation:

1. Move mouse to the start time (any available time slot).

    *UI*: The hovered slot is blue.

2. Click on the start time.

    *UI*: Change clicked slot to blue and show a popover above it (content: 'Choose reservation end time.' and '9:00-9:30').

3. Move mouse to the end time. 

    *UI*: The hovered slot and the start time slot are blue, the slots between are light blue. The time in the popover changes as slots are hovered.

4. Click on the end time.

    *UI*: Open a reservation confirmation modal.

5. Click 'Save' button.

    *UI*: Once request has been completed, close reservation modal and add the reservation info to the timeline.

## Part 1 - Selecting time

- [x] Select start time on first click.
  - `shared/availability-view/.../AvailabilityTimeline`
    - [x] Add handleReservationSlotClick function. Call `props.onReservationSlotClick({ resourceId: props.id, ...data })`.
    - [x] Add prop `id`, resource id.
  - `shared/availability-view/.../TimelineGroups`
    - [x] Add prop `onReservationSlotClick`.
    - [x] Pass `onReservationSlotClick={props.onReservationSlotClick}` to TimelineGroup.
  - `shared/availability-view/AvailabilityView`
    - [x] Add initial state: `{ selection: null }`.
    - [x] Add function handleReservationSlotClick. For now only calls `setState({ selection: { begin: ..., resourceId: ... } })`.
    - [x] Pass `onReservationSlotClick={this.handleReservationSlotClick}` to TimelineGroups.
- [x] Select end time on second click.
  - `shared/availability-view/AvailabilityView`
    - [x] Add prop `onSelect`.
    - [x] Modify handleReservationSlotClick. Add `if (this.state.selection) { ... }`:
      - If click data's resourceId is different or the click is before `selection.begin`, do nothing (i.e. prevent the click).
      - If click data's resourceId is the same and the click is the same or after `selection.begin`, update `state.selection.end` and call `this.props.onSelect(selection)` and clear `state.selection`.
- [x] Clear selection start when date changed.
  - Already done because AvailabilityView gets re-rendered when the day changes.

## Part 2 - Making a reservation

- [x] Open modal when end time selected.
  - `pages/admin-resources/AdminResourcesPage`
    - [x] Pass `onSelectionChange={this.handleSelect}` to AvailabilityView.
    - [x] Add function `handleSelect(selection)`.
[...]

PR #509, opened 2017-02-22
Highlighting

FelixLoether:
Refs #499.

### DIFF
--- a/app/shared/availability-view/AvailabilityView.js
+++ b/app/shared/availability-view/AvailabilityView.js
@@ -40,8 +40,7 @@ export default class AvailabilityView extends React.Component {
   }
 
   startSelection(slot) {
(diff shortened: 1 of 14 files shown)


lraivio:
- [ ] Selecting the end time does not seem to be working correctly. No matter what end time you select, the duration on the reservation form is always half an hour.

lraivio:
- [ ] Hovering mouse between the slots in causes some flickering. This could be prevented by creating the white spaces between slots using white borders instead of margins. The border radius and slot width will need to be adjusted accordingly. Something like:
```
width: 30px;
margin: 0;
border: 1px solid #fff;
border-radius: 3px;
```
Reservations would also need to be adjusted accordingly.

FelixLoether:
Fixes made.

PR #515, opened 2017-02-24
Prevent invalid selections

FelixLoether:
Refs #499.


FelixLoether:
- [ ] Use pointer as mouse on selectable slots.
- [ ] Rebase and make sure flickering doesn't happen.

Issue #506, opened 2017-02-20
Add reservation info modal to availability view


lraivio:
## Part 1 - Displaying info modal
- [x] Display info modal on availability view
- [x] Only show information that the reservation contains

## Part 2 - commenting, confirming and denying reservations
- [x] Move save comments button directly below comments
- [x] Block commenting past reservations
- [x] Add confirm button
- [x] Add deny button

## Part 3 - cancelling reservations
- [ ] Add cancel button

PR #531, opened 2017-03-06
Feature/506 reservation info modal to availability view - Part 2

lraivio:
This PR:
- Adds confirm button
- Adds deny buttons
- Updates comment button
- Prevents commenting past reservations

Refs #506.


app/state/reducers/ui/reservationInfoModalReducer.js
```diff
@@ -0,0 +1,36 @@
...
+
+    case types.UI.SHOW_RESERVATION_INFO_MODAL: {
+      return state.merge({
+        reservation: action.payload,
+        show: true,
+      });
+    }
+
+    case types.UI.HIDE_RESERVATION_INFO_MODAL: {
+      return initialState;
```
FelixLoether: - [ ] Currently when the modal is closed the modal becomes empty while it is disappearing. This could be fixed by not setting `reservation` to null here, i.e. changing this to `return state.merge({ show: false });`.

PR #559, opened 2017-03-23
Feature/548 add filters omat tilat

### LINKED ISSUE #548: Add type filters to "Omat tilat" admin view
Note, only type filters need to be added in this issue.

- [x] The type filters should function as toggles. Show a resource if its type is among the selected types. Filtering should be done in the frontend.
- [x] By default all types should be selected.
- [x] Only show types that at least one of the resources in the view has. The types can be get from resource data.
- [x] The types are not translated so only the Finnish name can be shown.

omat-tilat-filters

YoYuUm:
Closes #548


app/shared/resource-type-filter/ResourceTypeFilterContainer.js
```diff
@@ -32,7 +32,8 @@ export class UnconnectedResourceTypeFilterContainer extends Component {
 
   render() {
     return (
-      <div>
+      <div className="resource-type-filter-container">
+        <h6>Näytä</h6>
```
YoYuUm: I guess this could be translated, but I'm not actually sure if this title should be part of this component or actually part of its parent. What would be the english string?

app/shared/resource-type-filter/ResourceTypeFilterButton.js
```diff
@@ -0,0 +1,23 @@
...
+import Button from 'react-bootstrap/lib/Button';
+
+function ResourceTypeFilter({ active, onClick, resourceType }) {
+  return (
+    <Button
+      bsStyle={classnames({ primary: active, default: !active })}
+      className="resource-type-button"
+      onClick={() => onClick(resourceType)}
+    >
+      { resourceType }
```
lraivio: - [ ] You could also capitalize the resourceType text so all the texts in the buttons look the same.

app/shared/resource-type-filter/ResourceTypeFilterContainer.js
```diff
@@ -0,0 +1,49 @@
...
+            onClick={this.handleClick}
+            resourceType={resourceType}
+          />
+        )}
+      </div>
+    );
+  }
+}
+
+export default connect()(UnconnectedResourceTypeFilterContainer);
```
YoYuUm: The selectors and actions were here at the beginning, but I moved them to the AdminResourcesPage so we could reuse this component if needed. So I'll just return the component without "connect"

YoYuUm:
Fixed

lraivio:
- [ ] You can squash the fixups.

YoYuUm:
Squashed

PR #560, opened 2017-03-23
Add reservation length to reservationSlot

### LINKED ISSUE #557: Show reservation duration on tooltip when selecting end time
Color for the duration can be blue.

varaus_varauksen_kesto

YoYuUm:
Closes #557


YoYuUm:
Fixed and amended

PR #562, opened 2017-03-27
Style ResourceInfo to be aside resource images

### LINKED ISSUE #548: Add type filters to "Omat tilat" admin view

YoYuUm:
Closes #548


app/pages/resource/resource-info/resource-info.less
```diff
@@ -17,7 +17,25 @@
...
+  .details-row {
+    font-weight: 600;
+  }
+
+  .resource-equipment {
+    .label {
+      display: inline-block;
+      font-size: @font-size-small;
+      padding: 0.4em;
+      margin: 0.2em;
```
lraivio: - [ ] Probably better to not have left margin in the first element.

YoYuUm:
Fixed

lraivio:
- [ ] You can squash the fixup 👍

lraivio:
- [ ] Just noticed that the favorite button is not vertically aligned with the other content. You can fix that straight without a separate fixup commit.

Issue #572, opened 2017-03-30
Add recurring reservations


lraivio:
This feature should only be visible to admins. 

#### State
- [x] Add state for holding the required info for doing repeated reservations
  - baseReservation
  - finalDate
  - numberOfOccurrences
  - frequency
  - recurringReservations (so specific reservations can be removed if necessary)

#### RecurringTimeControls
- [x] Add select box for repeat options
  - none
  - daily
  - weekly
  - monthly
- [x] Add date-picker for selecting last day to repeat the event
- [x] Add number field to select the number of times the event should repeat
- [x] Keep the date-picker and the number field in sync
- [x] Add styling

#### Actions when clicking reserve
- [x] Add base recurring time to state

#### Reservation confirmation modal
- [x] Show recurringTimeControls
- [x] Populate recurring reservations in state based on user's choices
- [x] Show texts correctly in plural form if the reservation is recurring

#### Reservation success modal
- [x] Show all the successful reservations
- [x] Show all unsuccessful reservations (if needed)
- [x] Show reason for unsuccessful reservations (if needed)
- [x] Show successful and failed reservations ordered by time
- [x] Clear recurringReservations state after the modal is closed

#### Admin resources page
- [x] Add recurring reservation controls here as well

#### Mockup
![repeating-reservations](https://cloud.githubusercontent.com/assets/946697/24515249/ae2cf002-157f-11e7-8a54-c006b0cc00b9.png)

PR #586, opened 2017-04-18
Feature/572 recurring reservations base

lraivio:
First part of the recurring reservations feature

- Adds state and actions for recurring reservations
- Adds basic component for controlling recurring reservations
- Adds the recurring reservation controls to resource page

Refs #572.


app/shared/recurring-reservation-controls/RecurringReservationControls.js
```diff
@@ -0,0 +1,60 @@
...
+import Select from 'react-select';
+
+import { injectT } from 'i18n';
+
+function createOptionRenderer(t) {
+  return option => t(option.label);
+}
+
+function createValueRenderer(t) {
+  return option => t(option.label);
```
YoYuUm: - [ ] Probably you meant t(option.value)

lraivio: Actually the `option.label` is correct there as well. The valueRenderer tells how to display the selected value in the input. But good that you noticed.

PR #588, opened 2017-04-18
Improve karma test configs

lraivio:
Shows diffs of expected and actual values on failed tests.


YoYuUm:
Oh this is so nice. No more annoying console.logs :)

PR #587, opened 2017-04-18
Feature/572 recurring reservations api calls

lraivio:
This PR:
- Adds recurring reservations to state
- Displays recurring reservations on reservation confirm modal
- Makes actual API calls for every recurring reservation.

Refs #572.

PR #589, opened 2017-04-18
Feature/572 recurring reservations failed reservations

lraivio:
This PR:
- Adds failed reservations to state
- Shows failed reservations on reservation success modal below the successful reservations
- Shows fail reasons below every failed reservation

Built on top of #587 so that should be merged first.

Refs #572.

### DIFF
--- a/app/state/reducers/ui/reservationsReducer.js
+++ b/app/state/reducers/ui/reservationsReducer.js
@@ -1,5 +1,5 @@
-import without from 'lodash/without';
 import includes from 'lodash/includes';
+import without from 'lodash/without';
 import Immutable from 'seamless-immutable';
 
 import types from 'constants/ActionTypes';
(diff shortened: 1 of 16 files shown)


app/state/reducers/ui/reservationsReducer.js
```diff
@@ -29,6 +29,15 @@ function selectReservationToEdit(state, action) {
   });
 }
 
+function parseError(error) {
+  if (error.response && error.response.non_field_errors && error.response.non_field_errors.length) {
+    return error.response.non_field_errors.join('. ').replace('[\'', '').replace('\']', '');
+  } else if (error.response && error.response.detail) {
+    return error.response.detail;
+  }
+  return 'Jotain meni vikaan';
```
YoYuUm: This message is not translated but I guess it is fine, as the backend messages are in finnish too?

YoYuUm:
I've approved this PR accidentally, but it still needs a rebase. After it is done it is good to go.

app/state/reducers/ui/reservationsReducer.js
lraivio: Yeah, I think it's okay to have that just in Finnish for the moment. The API error messages are also in Finnish and these are only shown to admins. If we at some point make recurring reservations available to regular users as well, then we might translate these as well.

lraivio:
Alright, thanks @YoYuUm. Rebase is now done.

PR #595, opened 2017-04-21
Feature/572 recurring reservations final polishes

lraivio:
Final polishes to recurring reservations.

Refs #572.

### DIFF
--- a/app/state/recurringReservations.js
+++ b/app/state/recurringReservations.js
@@ -1,6 +1,9 @@
 import moment from 'moment';
 import { createAction, handleActions } from 'redux-actions';
 
+import actionTypes from 'constants/ActionTypes';
+import modalTypes from 'constants/ModalTypes';
+
 // actions
 // -------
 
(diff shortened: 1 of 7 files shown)


app/state/recurringReservations.js
```diff
@@ -89,6 +92,13 @@ const recurringReservationsReducer = handleActions({
   [actions.changeNumberOfOccurrences]: (state, action) => adjustState({
     ...state, numberOfOccurrences: parseInt(action.payload, 10),
   }),
+  [actionTypes.UI.CLOSE_MODAL]: (state, action) => {
+    const modalType = action.payload;
+    if (modalType === modalTypes.RESERVATION_SUCCESS) {
+      return initialState;
+    }
+    return state;
+  },
```
YoYuUm: Good point.

- [ ] I think this should happen when RESERVATION_CONFIRM modal type gets closed as well.

lraivio: Yeah, I was also thinking about whether it should or not. But maybe it makes more sense to reset the state also after the confirm modal is closed. I made the fix and also rebased already.

lraivio:
@YoYuUm: fixes made.

PR #608, opened 2017-04-28
Improve availability view expand icon

### LINKED ISSUE #607: Make availability view datepicker expand icon also close the calendar

lraivio:
A quick hotfix before we get the new day picker to the availability view.
The default date-picker calendar icon works well in closing and opening
the calendar. This commit makes that icon invisible and adds an expand icon
on top of that. The icon direction changes when the calendar is open to make
it cleared that the icon can be used to close the calendar.
Closes #607.

### DIFF
--- a/app/shared/availability-view/availability-view.less
+++ b/app/shared/availability-view/availability-view.less
@@ -86,14 +86,51 @@
       z-index: 3;
     }
     .react-date-field__calendar-icon {
-      display: none;
+      z-index: 10;
+      border: none;
+      &:hover {
+        background-color: transparent;
+        .react-date-field__calendar-icon-inner {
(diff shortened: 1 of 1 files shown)


app/shared/availability-view/availability-view.less
```diff
@@ -86,14 +86,51 @@
...
+        background-color: transparent;
+
+        &::after {
+          font-family: 'Glyphicons Halflings';
+          content: "\e252";
+          visibility: visible;
+          color: white;
+          font-size: 18px;
+          position: absolute;
+          right: 0;
```
YoYuUm: - [ ] You could add about 20px of right, so the icon is not that far way from the input.

PR #612, opened 2017-05-05
Allow to remove times from recurring reservation

### LINKED ISSUE #571: Allow user to remove specific times from recurring reservations
If user is doing recurring reservations, she should be able to remove some reservation times before saving the reservations.

remove-reservations-before-confirming

YoYuUm:
Closes #571

I've updated the numberOfOcurrencesLabel english label so it's easier to understand. Should I change the finnish and swedish labels?

### DIFF
--- a/app/i18n/messages/en.json
+++ b/app/i18n/messages/en.json
@@ -105,7 +105,7 @@
   "RecurringReservationControls.frequencyNone": "None",
   "RecurringReservationControls.frequencyWeekly": "Weekly",
   "RecurringReservationControls.lastTimeLabel": "Ends on",
-  "RecurringReservationControls.numberOfOccurrencesLabel": "Occurrences",
+  "RecurringReservationControls.numberOfOccurrencesLabel": "Repetitions",
   "ReservationAccessCode.defaultText": "PIN code",
   "ReservationCalendar.header": "Select a day",
   "ReservationCalendarControls.confirmChanges": "Confirm the changes",
--- a/app/shared/compact-reservation-list/CompactReservationList.js
+++ b/app/shared/compact-reservation-list/CompactReservationList.js
@@ -1,14 +1,14 @@
 import React, { Component, PropTypes } from 'react';
+import Glyphicon from 'react-bootstrap/lib/Glyphicon';
 
 import TimeRange from 'shared/time-range';
 
 class CompactReservationList extends Component {
-  constructor(props) {
-    super(props);
-    this.renderReservation = this.renderReservation.bind(this);
-  }
+  renderFixedReservation = reservation => this.renderReservation(reservation);
 
-  renderReservation(reservation) {
+  renderRemovableReservation = reservation => this.renderReservation(reservation, true);
+
+  renderReservation = (reservation, removable = false) => {
     let resourceName = null;
     if (this.props.resources) {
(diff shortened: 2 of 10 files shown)


app/shared/compact-reservation-list/CompactReservationList.spec.js
```diff
@@ -66,4 +68,28 @@ describe('shared/compact-reservation-list/CompactReservationList', () => {
...
+      const props = { reservations, removableReservations };
+      const buttons = getWrapper(props).find({ glyph: 'remove-circle' });
+      expect(buttons).to.have.length(2);
+    });
+
+    it('passes onRemoveClick func to onClick prop', () => {
+      const removableReservations = [Reservation.build()];
+      const props = { removableReservations };
+      const button = getWrapper(props).find({ glyph: 'remove-circle' });
+      expect(defaultProps.onRemoveClick.callCount).to.equal(0);
```
lraivio: - [ ] Whenever we are testing functions with mocks, let's try to create the mock on the test that is using it. Trouble with mocks in defaultProps is that they can be used or called in multiple tests which can break some tests if we test callCounts and such. 

In this case you can just create the `onRemoveClick` mock in this test and pass it as a prop.

app/shared/compact-reservation-list/CompactReservationList.spec.js
```diff
@@ -18,6 +19,7 @@ describe('shared/compact-reservation-list/CompactReservationList', () => {
     resources: Immutable({
       [resource.id]: resource,
     }),
+    onRemoveClick: simple.mock(),
```
lraivio: - [ ] Change this to just plain arrow function so we don't accidentally use the same mock in multiple tests.
```
onRemoveClick: () => null,
```

app/i18n/messages/en.json
```diff
@@ -105,7 +105,7 @@
   "RecurringReservationControls.frequencyNone": "None",
   "RecurringReservationControls.frequencyWeekly": "Weekly",
   "RecurringReservationControls.lastTimeLabel": "Ends on",
-  "RecurringReservationControls.numberOfOccurrencesLabel": "Occurrences",
+  "RecurringReservationControls.numberOfOccurrencesLabel": "Repeatitions",
```
lraivio: - [ ] Should be "Repetitions".

lraivio: The Finnish and Swedish translations should be alright.

app/shared/compact-reservation-list/CompactReservationList.js
```diff
@@ -18,8 +19,16 @@ class CompactReservationList extends Component {
       <li key={reservation.begin}>
         {resourceName ? `${resourceName}: ` : ''}
         <TimeRange begin={reservation.begin} end={reservation.end} />
+        {removable &&
+          <Glyphicon
+            glyph="remove-circle"
+            onClick={() => this.props.onRemoveClick(reservation.begin)}
```
YoYuUm: I can't get the reservation from the props, so even if there would be a handleRemoveClick function, i would need to invoke it from the component like {() => this.handleRemoveClick(reservation)}

lraivio:
Nice work :+1:
You can squash the fixups.

PR #614, opened 2017-05-09
Hotfix/613 recurring reservation control improvements

### LINKED ISSUE #613: Recurring reservation control improvements on some browsers
- [x] Calendar icon is not properly positioned on Edge and Firefox
- [x] Investigate whether it would be easily possible to get arrows to number of repetition field (native number input ?) on IE11 and Edge.

recurring-reservations-on-ie

YoYuUm:
closes #613.

It uses a polyfill to solve the problem with the numeric input.

I'ld see convenient to add a maximum repetitions. Now the browser can freeze if the user types in a large number.

### DIFF
--- a/app/shared/date-picker/date-picker.less
+++ b/app/shared/date-picker/date-picker.less
@@ -1,5 +1,9 @@
 .date-picker {
-
+  input {
+    // Width value won't be taken in account because date-picker display is flex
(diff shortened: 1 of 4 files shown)


lraivio:
I added issue #615 for limiting the number of recurring reservations.

PR #625, opened 2017-05-19
Feature/621 show user location

### LINKED ISSUE #621: Map - show users current location
- [x] Ask user's current location
- [x] Show user's current location on the map

YoYuUm:
Closes #621


app/shared/resource-map/MapContainer.js
```diff
@@ -76,4 +87,9 @@ UnconnectedResourceMapContainer.propTypes = {
   }),
 };
 
-export default connect(selector)(UnconnectedResourceMapContainer);
+UnconnectedResourceMapContainer.propTypes = {
+  ...UnconnectedResourceMapContainer.propTypes,
+  ...geoPropTypes,
```
lraivio: - [ ] Why not have these just in the same place where the other UnconnectedResourceMapContainer propTypes are declared?

app/shared/resource-map/UserMarker.js
```diff
@@ -0,0 +1,35 @@
...
+    return (
+      <LeafletMarker
+        icon={icon}
+        position={[this.props.latitude, this.props.longitude]}
+      />
+    );
+  }
+}
+
+UserMarker.propTypes = {
```
lraivio: - [ ] We could start declaring the propTypes as static propTypes or before a stateless functional component. So let's move towards that whenever we add new files or refactor old ones.

lraivio:
You can squash the fixups 👍

PR #633, opened 2017-05-25
Feature/628 expand map

### LINKED ISSUE #628: Expand map to be cover the whole search page
- [x] When list view is selected, the map should be in the background and user should not be able to interact with it. The map should be faded out / have a white overlay on top of it.
- [x] When map view is selected the list should move down out of the way (maybe disappear completely) and user should be able to interact with the map.
- [x] Search controls should always be visible and on top of the map.
- [x] The list/map toggle button should always be visible and on top of the map. 

Note: The desktop sidebar will be implemented later.

search-page-list-mobile

search-page-map-mobile

search-page-list-desktop

search-page-map-resource-selected

YoYuUm:
Closes #628 

It doesn't implement the resources down animation because I think it will be easier to implement later on, when we get rid of the footer.

### DIFF
--- a/app/shared/resource-map/MapContainer.spec.js
+++ b/app/shared/resource-map/MapContainer.spec.js
@@ -28,6 +28,16 @@ describe('shared/resource-map/MapContainer', () => {
     expect(map).to.have.length(1);
   });
 
+  it('adds showMap classname to wrapper div if showMap prop is passed', () => {
+    const wrapper = getWrapper();
+    expect(wrapper.prop('className')).to.contain('app-ResourceMap__showMap');
+  });
+
+  it('does not add showMap classname to wrapper div if showMap prop is not passed', () => {
+    const wrapper = getWrapper({ showMap: false });
+    expect(wrapper.prop('className')).to.not.contain('app-ResourceMap__showMap');
+  });
+
   it('Map is centered at default position', () => {
     const defaultPosition = [60.372465778991284, 24.818115234375004];
     const map = getWrapper().find(Map);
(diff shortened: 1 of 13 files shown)


app/shared/resource-map/MapContainer.spec.js
```diff
@@ -65,13 +75,8 @@ describe('shared/resource-map/MapContainer', () => {
     expect(element).to.have.length(0);
   });
 
-  it('does not render an overlay div if showMap prop is true', () => {
+  it('renders an overlay div if showMap prop is true', () => {
```
lraivio: - [ ] The overlay is always rendered, not just when the showMap prop is true.

app/shared/resource-map/MapContainer.spec.js
```diff
@@ -27,6 +28,16 @@ describe('shared/resource-map/MapContainer', () => {
     expect(map).to.have.length(1);
   });
 
+  it('adds showMap classname to wrapper div if showMap prop is passed', {
+    const wrapper = getWrapper()
```
lraivio: - [ ] Semicolon missing.

lraivio:
- [ ] There is an `Uncaught (in promise) TypeError: Cannot read property 'location' of undefined` -error when reloading the search page when something is already searched. 
For example `http://localhost:3000/search?date=2017-05-26&people=&purpose=&search=sali`.

YoYuUm:
This is not compatible with IE 10 because of the css transition.

It could be compatible adding a showOverlay attribute to the state:
```
  componentWillReceiveProps(nextProps) {
    if (this.props.showMap === false && nextProps.showMap) {
      window.setTimeout(() => {
        this.setState({
          showOverlay: false,
        });
      }, 400);
    }
    if (this.props.showMap && nextProps.showMap === false) {
      this.setState({
        showOverlay: true,
      });
    }
  }
```

And then rendering the overlay layer if state.showOverlay is true:
```
{this.state.showOverlay && <div className="app-ResourceMap__overlay" />}
```

This solution trade-off is that the background transition from map to list is lost.

YoYuUm:
Fixed, the tests issues have been already squashed in 994c2f9

lraivio:
Looks good now :+1:

Issue #631, opened 2017-05-22
Update search control styles


lraivio:
- [x] textbox
- [x] Datepicker button
- [x] Number of people button
- [x] Purpose select

PR #636, opened 2017-05-26
Feature/631 search controls styles part 1

lraivio:
First part of search controls style updates.
Updates search box and date picker styles. Adds first version of sass mixin to add rounded button styles.

Rest of the updates will be done in part 2.

Refs #631.

### DIFF
--- a/app/pages/search/SearchPage.js
+++ b/app/pages/search/SearchPage.js
@@ -74,7 +74,6 @@ class UnconnectedSearchPage extends Component {
           showMap={showMap}
         />
         <div className="app-SearchPage__content">
-          <h1>{t('SearchPage.title')}</h1>
           <SearchControls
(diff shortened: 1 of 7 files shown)


app/pages/search/SearchPage.js
```diff
@@ -74,7 +74,6 @@ class UnconnectedSearchPage extends Component {
           showMap={showMap}
         />
         <div className="app-SearchPage__content">
-          <h1>{t('SearchPage.title')}</h1>
```
YoYuUm: - [ ] Perhaps the translations for SearchPage.title show be removed too?

lraivio: It's still needed in the page title (in DocumentTitle component). But yeah, otherwise it would be best to remove it as well.

PR #642, opened 2017-05-30
Feature/630 update search results styles

### LINKED ISSUE #630: Update search results styles
Note: The available times will be implemented in another issue.

search-page-list-mobile

YoYuUm:
Closes #630


app/shared/resource-list/ResourceListItem.js
```diff
@@ -26,7 +28,20 @@ class ResourceListItem extends Component {
             height={420}
             image={getMainImage(resource.images)}
             width={700}
-          />
+          >
+            <Label
+              className="app-ResourceListItem__peopleCapacity"
+              shape="rounded"
```
lraivio: - [ ] Here shape "circle" could be used so there would be less need for style customizations.

app/shared/resource-list/ResourceListItem.js
```diff
@@ -26,7 +28,20 @@ class ResourceListItem extends Component {
...
+              size="mini"
```
lraivio: - [ ] Size "medium" would require less style customizations.

app/shared/resource-list/ResourceListItem.js
```diff
@@ -26,7 +28,20 @@ class ResourceListItem extends Component {
...
+              theme="orange"
+            >
+              {resource.peopleCapacity}
+              <FontAwesome name="user" />
```
lraivio: - [ ] We could maybe use the icon "users" for people capacity even though it differs from the designs. But would represent number of people better imo.

lraivio:
- [ ] The ResourceAvailability could have rounded corners similar to what the new available time buttons would have. The text color could also be white.

image

image

lraivio:
- [ ] The ResourceAvailability could be positioned to bottom left beside the picture like the first green button in the image below.

image

lraivio:
- [ ] The search results list could have a max-width so the results won't expand too much and there would be some space for a side panel in the future. Maybe something like 800px.
- [ ] The image should not expand much at all.

image

lraivio:
- [ ] The search results should be aligned with the other elements on the search page.

image

YoYuUm:
Fixed

lraivio:
Nice :+1:

PR #646, opened 2017-06-01
Feature/629 basic resource page

### LINKED ISSUE #629: Implement basic resource page styles
- [x] Narrow the page on desktop to resemble the mobile layout (The page will later be displayed on the side of the search page)
- [x] Make the picture full page width
- [x] Header
- [x] Icons over the picture
- [x] Resource description
- [x] Reservation info
- [x] Equipment

YoYuUm:
Closes #629 

- [ ] Missing check espoo custom styles

### DIFF
--- a/app/pages/resource/reservation-calendar/ReservationCalendarContainer.js
+++ b/app/pages/resource/reservation-calendar/ReservationCalendarContainer.js
@@ -1,7 +1,5 @@
 import React, { Component, PropTypes } from 'react';
 import { connect } from 'react-redux';
-import Col from 'react-bootstrap/lib/Col';
-import Row from 'react-bootstrap/lib/Row';
 import { browserHistory } from 'react-router';
 import { bindActionCreators } from 'redux';
 
@@ -159,21 +157,11 @@ export class UnconnectedReservationCalendarContainer extends Component {
 
   render() {
     return (
-      this.props.isAdmin ?
-        <div className="reservation-calendar">
-          <h3 id="resource-calendar-header">{this.props.t('ReservationCalendar.header')}</h3>
-          {this.renderCalendar()}
-          {this.renderTimeSlots()}
-        </div> :
-          <Row className="reservation-calendar">
-            <Col sm={4} xs={12}>
-              <h3 id="resource-calendar-header">{this.props.t('ReservationCalendar.header')}</h3>
-              {this.renderCalendar()}
-            </Col>
-            <Col sm={8} xs={12}>
-              {this.renderTimeSlots()}
-            </Col>
-          </Row>
+      <div className="reservation-calendar">
+        <h3 id="resource-calendar-header">{this.props.t('ReservationCalendar.header')}</h3>
+        {this.renderCalendar()}
+        {this.renderTimeSlots()}
+      </div>
     );
   }
 }
--- a/app/pages/resource/resource-info/ResourceInfo.spec.js
+++ b/app/pages/resource/resource-info/ResourceInfo.spec.js
@@ -81,7 +81,7 @@ describe('pages/resource/resource-info/ResourceInfo', () => {
     );
 
     expect(peopleCapacityLabel.is(Label)).to.be.true;
-    expect(peopleCapacityLabel.prop('shape')).to.equal('circle');
+    expect(peopleCapacityLabel.prop('shape')).to.equal('rounded');
     expect(peopleCapacityLabel.prop('size')).to.equal('medium');
     expect(peopleCapacityLabel.prop('theme')).to.equal('orange');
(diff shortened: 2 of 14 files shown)


app/pages/resource/resource-info/ResourceInfo.spec.js
```diff
@@ -55,16 +60,48 @@ describe('pages/resource/resource-info/ResourceInfo', () => {
...
+  it('renders a label with people capacity', () => {
+    const peopleCapacityLabel = getWrapper().find(
+      '.app-ResourceInfo__peopleCapacity'
+    );
+
+    expect(peopleCapacityLabel.is(Label)).to.be.true;
+    expect(peopleCapacityLabel.prop('shape')).to.equal('circle');
+    expect(peopleCapacityLabel.prop('size')).to.equal('medium');
+    expect(peopleCapacityLabel.prop('theme')).to.equal('orange');
+    expect(peopleCapacityLabel.children().at(0).text()).to.equal('16');
```
lraivio: - [ ] Having tests that rely this deeply on the html structure of the component is a bit problematic. Just removing the space between the icon and text will break the tests. Or switching the icon and text around.
I think it's enough to check that the value we are interested get's rendered. So here would be enough to check that label children contain resource.peopleCapacity. The hardcoded value '16' is also a bit problematic. Is the value always '16' or just because the defaultProps.resource.peopleCapacity is 16. The test doesn't really tell that. 

So testing the children could be something like:
```javascript
const expectedValue = defaultProps.resource.peopleCapacity;
expect(peopleCapacityLabel.children().contains(expectedValue)).to.be.true;
```

app/pages/resource/resource-info/ResourceInfo.spec.js
```diff
@@ -55,16 +60,48 @@ describe('pages/resource/resource-info/ResourceInfo', () => {
...
+    const typeLabel = getWrapper().find(
+      '.app-ResourceInfo__type'
+    );
 
-    expect(resourceIcons.length).to.equal(1);
+    expect(typeLabel.is(Label)).to.be.true;
+    expect(typeLabel.prop('shape')).to.equal('rounded');
+    expect(typeLabel.prop('size')).to.equal('medium');
+    expect(typeLabel.prop('theme')).to.equal('blue');
+    expect(typeLabel.children().at(0).is(FontAwesome)).to.be.true;
```
lraivio: - [ ] Same thing here.

app/pages/resource/reservation-calendar/ReservationCalendarContainer.js
```diff
@@ -166,11 +166,11 @@ export class UnconnectedReservationCalendarContainer extends Component {
           {this.renderTimeSlots()}
         </div> :
           <Row className="reservation-calendar">
-            <Col sm={4} xs={12}>
+            <Col sm={12}>
```
lraivio: - [ ] We can now get rid of the Rows and Cols altogether.

lraivio:
- [ ] There could be a bit more margin between the resource and unit names.

image

lraivio:
- [ ] The price text should not be all uppercase.

image

lraivio:
- [ ] Have the same horizontal margin in the resource info top bar, bottom bar and resource description text. Now the top bar has 10px, resource description has 15px and the bottom bar is not aligned to left at all. 20px would probably be a good margin.

image

lraivio:
Looks already very good. Let's do couple of changes still. The previous fixes you can already squash.

lraivio:
- [ ] Move "users" icon before the number in people capacity label. More consistent with other labels that way even though it's differently in the designs.

lraivio:
- [ ] Have the price info aligned to left and all the labels aligned to right with 20px margin between them.

image

lraivio:
- [ ] The price info could be displayed with a larger font-size and the resource name with slightly smaller font-size.

lraivio:
- [ ] If for some reason the image can't be fetched, the styles break. This should not happen in production but still good to put min-height for the image and maybe a background color as well.

image

YoYuUm:
@lraivio Fixed

lraivio:
Very nice :+1: 
Still couple of small fixes and then this is ready.

- [ ] When people capacity is high the number of people does not fit into the label. Let's use a shape="rounded" label to fix this.

image

image

- [ ] For the missing image styles let's use `min-height: 400px` and `background-color: $medium-gray` to make it look like there should be an image there.

image

lraivio:
You can make these changes straight away to the previous commits.

YoYuUm:
Fixed

lraivio:
Nice job :+1:

PR #665, opened 2017-06-21
Feature/660 show resources as card

### LINKED ISSUE #660: Map - Show resources in the selected unit as cards
- [x] Highlight selected unit (map marker).
- [x] Separate the "resource card" code from resource list so it can be used both in search results and here as cards.  
- [x] Add cards to bottom of the page.
- [x] Add arrow to switch the topmost resource.
- [x] Make the card a link to the resource page.

image

YoYuUm:
closes #660


app/shared/resource-map/MapContainer.js
```diff
@@ -93,4 +102,12 @@ export class UnconnectedResourceMapContainer extends React.Component {
   }
 }
 
-export default connect(selector)(geolocated()(UnconnectedResourceMapContainer));
+function mapDispatchToProps(dispatch) {
```
lraivio: - [ ] This was the old way of giving dispatch to actions in Varaamo. Since this is a new container you can just give the connect the actions object. 
```javascript
const actions = {
  selectUnit,
};

export default connect(selector, actions)(geolocated()(UnconnectedResourceMapContainer));
```

That way the selectUnit will be just a function in the props. A lot nicer to use and mock in tests.

app/shared/resource-map/Marker.spec.js
```diff
@@ -12,29 +12,31 @@ function getWrapper(props) {
...
 }
 
 describe('shared/resource-map/Marker', () => {
   describe('handleClick', () => {
-    let browserHistoryMock;
+    let selectUnitMock;
 
     before(() => {
-      browserHistoryMock = simple.mock(browserHistory, 'push');
+      selectUnitMock = simple.mock(browserHistory, 'push');
```
lraivio: - [ ] browserHistory no longer needs to be mocked here, right?

app/pages/search/results/SearchResults.spec.js
```diff
@@ -75,6 +76,19 @@ describe('pages/search/results/SearchResults', () => {
         expect(toggleButton).to.have.length(1);
         expect(toggleButton.text()).to.contain('SearchResults.showList');
       });
+
+      describe('whit selectedUnitId', () => {
```
lraivio: - [ ] "whit" -> "with".

app/assets/styles/_application-variables.scss
```diff
@@ -55,3 +55,7 @@ $element-size-small:           24px;
 $element-size-mini:            18px;
 $button-border-radius:         30px;
 $button-horizontal-padding:    15px;
+
+// Component vairables
```
lraivio: - [ ] "vairables" -> "variables".

app/shared/resource-map/MapContainer.js
```diff
@@ -72,6 +76,7 @@ export class UnconnectedResourceMapContainer extends React.Component {
         <Map
           center={center}
           className="app-ResourceMap__map"
+          onClick={this.handleOnClick}
```
lraivio: - [ ] Here you can just use the `this.props.actions.searchMapClick`. No need for a separate handle method. If there was a need the method should be named just `handleClick` or `handleMapClick` this case.

YoYuUm:
Fixed

PR #728, opened 2017-07-27
Feature/725 search page improvements

### LINKED ISSUE #725: Search page improvements
- [x] Add address to resource cards
- [x] Add map icon/marker to show map button. A list icon to show list button.
- [x] Show stacked cards when one unit

YoYuUm:
Closes #725

### DIFF
--- a/app/pages/search/results/MapToggle.spec.js
+++ b/app/pages/search/results/MapToggle.spec.js
@@ -1,5 +1,6 @@
 import { expect } from 'chai';
 import React from 'react';
+import Glyphicon from 'react-bootstrap/lib/Glyphicon';
 
 import { shallowWithIntl } from 'utils/testUtils';
 import MapToggle from './MapToggle';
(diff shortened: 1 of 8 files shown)


lraivio:
- [ ] One nice usability upgrade would be to always have both of the side arrows in the card stack visible. That way browsing the cards would be faster.

image

->

image

app/pages/search/results/MapToggle.spec.js
```diff
@@ -45,4 +46,18 @@ describe('pages/search/results/MapToggle', () => {
       expect(getButtonText(true)).to.contain('MapToggle.showList');
     });
   });
+
+  describe('button text', () => {
+    function getButtonIcon(mapVisible) {
+      return getWrapper({ mapVisible }).find('.app-MapToggle__icon').find(Glyphicon);
+    }
+
+    it('renders show map text if map is not visible', () => {
```
lraivio: - [ ] This could read "icon" instead of "text".

app/pages/search/results/MapToggle.spec.js
```diff
@@ -45,4 +46,18 @@ describe('pages/search/results/MapToggle', () => {
...
+      expect(getButtonIcon(false).prop('glyph')).to.equal('map-marker');
+    });
+
+    it('renders show list text if map is visible', () => {
```
lraivio: - [ ] This could read "icon" instead of "text".

YoYuUm:
@lraivio So do you mean that the arrow would loop from last resource to first oneeif you keep clicking the right arrow? (and from first to last if you click the left arrow)

lraivio:
@YoYuUm Yes exactly. Can be in another PR if it requires a lot of extra work.

YoYuUm:
Fixed. The test issues have been squashed already

PR #730, opened 2017-07-31
Feature/724 sidebar static on desktop

### LINKED ISSUE #724: Make sidebar navigation static on desktop
- [x] Make sidebar navigation static on desktop. (Should be always visible). On mobile and tablet devices should work the same way as now.
- [x] Make "Lisätietoa palvelusta" navigation link easier to find by putting it under "search" link and using same link styles than the rest of the navigation links.

YoYuUm:
Closes #724


app/i18n/messages/fi.json
```diff
@@ -80,6 +79,7 @@
   "MapToggle.showList": "Näytä listaus",
   "MapToggle.showMap": "Näytä kartalla",
   "MiniModal.buttonText": "Valmis",
+  "Navbar.aboutLink": "Lisätietoa tietoa",
```
lraivio: - [ ] This should be "Lisää tietoa".

lraivio:
- [ ] The cursor could maybe be default on Desktop over the sidenavbar toggle since it can't be clicked.

YoYuUm:
Fixed

PR #739, opened 2017-08-07
Feature/727 resource map to app container

### LINKED ISSUE #727: Add map to resource page
Move the map used in search page app to AppContainer so it the same map can be used in the background of both the search page and the resource page. 
This should also at least partly solve #678.

YoYuUm:
Closes #727


YoYuUm:
Fixed callback() misscall error.

PR #747, opened 2017-08-11
Add travis and codecov

hukka:


hukka:
If everything works fine, we can remove circle later (assuming that it's wasted effort to maintain configs for both CIs).

Issue #672, opened 2017-06-23
Make sure Espoo styles still work nicely with the new UI


lraivio:
- [x] Sidebar colors
- [x] Font
- [x] Logo

lraivio:
Closed in #775.

PR #758, opened 2017-10-02
Feature/672 espoo styles - part 1

lraivio:
This PR:
- Updates espoo navbar styles.
- Makes sure espoo font is used everywhere.
- Removes old unneeded styles.

Refs #672.


app/assets/styles/customization/espoo/customization.scss
```diff
@@ -29,25 +29,34 @@
...
-      .open > a:focus {
-        background-color: darken($blue, 6.5%);
-      }
+  .app-SideNavbar__toggle {
+    background-color: $blue;
+    color: $white;
+
+    &:hover, &:focus {
+      color: $white;
     }
```
YoYuUm: This should also have styles for anchor colors, so "Close" button is white. Currently it has the same color than the background

lraivio:
Fixed.

PR #760, opened 2017-10-04
Feature/752 add distance slider

### LINKED ISSUE #752: Add distance filter to search page
It's enough to add the filter without the on/off toggle in this issue. The toggle will be added later.

tilan-listaus-etaisyys_sidebyside

YoYuUm:
Closes #752


lraivio:
- [ ] The upper limit could be 20 km.

lraivio:
- [ ] If the toggle is off, rather make the slider disabled than remove it altogether. That way it should more intuitive for the users at start.

lraivio:
- [ ] You can add a tooltip to the slider easily with [createSliderWithTooltip](http://react-component.github.io/slider/#createsliderwithtooltip(slider-|-range)-=>-react.component).

That could replace the current Badge in the modal.

YoYuUm:
fixed

lraivio:
- [ ] Make the distance filter button similar to the other buttons.
So smaller icon and header "distance" beside it.
Then the currently selected value under them.
If it's off the value could be just "off"
If there is no upper limit then it can be "any" or something like that.

image

lraivio:
- [ ] The modal header should be "Distance"
- [ ] Maybe to make it easier for the users to understand the toggle and geolocation we could have it inside the modal and have a text like "Use your current location to filter by distance". Or something like that.
- [ ] I know we removed the label that showed the currently selected distance, but we could still have the current value shown as text above the slider.
Maybe something like "Max distance: 12 km"

image

PR #770, opened 2017-11-30
Vantaa customizations

vrpl:


vrpl:
Travis CI failed because: 30 11 2017 12:36:14.134:ERROR [launcher]: Cannot start Chrome. I don't have the option to restart the build. Could someone with write permissions try to do that?

PR #796, opened 2018-08-24
Fixed CSS especially for mobile devices, and added footer link to old Varaamo.

tikkari2:


tikkari2:
This is replaced with two other pull requests.

PR #807, opened 2018-11-13
Changed versions as suggested by Github checker.

tikkari2:
Changed versions as suggested by Github security checker.


tikkari2:
Changes included in PR 808. This is not required.

Issue #817, opened 2018-11-27
Change naming of external calendar url to external reservation url


turtommy:
Naming of field external_calendar_url will be changed to external_reservation_url in Respa backend. Corresponding changes must be done to Varaamo also.

More info:
https://github.com/City-of-Helsinki/respa/pull/389#discussion_r235515331

PR #818, opened 2018-11-28
Renamed external calendar url to external reservation url

turtommy:
Renamed the external calendar url field to correspond changes in Respa.


turtommy:
Related to issue #817

PR #837, opened 2019-01-08
Vertically align time slots in reservation calendar

metrip:
- To align times, add a placeholder component to reservation calendar day columns with start time later than earliest start time
- Manage vertical offset for "mobile" 3 day view 
- Take into account days when unit is closed (in time slots, start field: date only, no end field)

### DIFF
--- a/app/pages/resource/reservation-calendar/time-slots/TimeSlots.js
+++ b/app/pages/resource/reservation-calendar/time-slots/TimeSlots.js
@@ -2,10 +2,13 @@ import React, { Component, PropTypes } from 'react';
 import Loader from 'react-loader';
 import moment from 'moment';
 import classnames from 'classnames';
+import findIndex from 'lodash/findIndex';
+import minBy from 'lodash/minBy';
 
 import constants from 'constants/AppConstants';
 import { injectT } from 'i18n';
 import TimeSlot from './TimeSlot';
+import TimeSlotPlaceholder from './TimeSlotPlaceholder';
 import utils from '../utils';
 
 class TimeSlots extends Component {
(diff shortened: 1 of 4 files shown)


app/pages/resource/reservation-calendar/time-slots/TimeSlots.js
```diff
@@ -24,30 +27,85 @@ class TimeSlots extends Component {
...
+    const timeSlotPlaceholderSizes = slots.map((slot) => {
+      if (!slot[0] || !slot[0].end) {
+        return null;
+      }
+      const currentStart = moment(slot[0].start).set(dateForTimeComparison);
+      return currentStart.diff(earliestStart, 'minutes') / 30;
+    });
+
+    const selectedDateIndex = findIndex(
+      slots,
```
vharvala:
```suggestion
      firstTimeSlots,
```

app/pages/resource/reservation-calendar/time-slots/TimeSlots.js
```diff
@@ -24,30 +27,85 @@ class TimeSlots extends Component {
...
+      slot => moment(slot[0].start).format(constants.DATE_FORMAT) === selectedDate
```
vharvala: When opening a resource from search result list, an error page flashes quickly and then it continues to the resource page normally. It happens because of this row's `slot[0].start` since slots array is not fully populated on first render. It needs an additional check or maybe using `firstTimeSlots` instead since it has prechecked values.

```suggestion
      slot => moment(slot.start).format(constants.DATE_FORMAT) === selectedDate
```

app/pages/resource/reservation-calendar/time-slots/TimeSlots.js
metrip: Has to be done over `slots` since we are finding an index and `firstTimeSlots` is filtered for days when the unit is closed (no `end` field).

app/pages/resource/reservation-calendar/time-slots/TimeSlots.js
metrip: Will check

app/pages/resource/reservation-calendar/time-slots/TimeSlots.js
vharvala: Ah yes!

PR #854, opened 2019-02-11
[Fix] Upgrade react-router

Nizar-Rahme:


Nizar-Rahme:
[Replace react-router with the latest version of react-router-dom](https://trello.com/c/wDVuVRZa/26-replace-react-router-with-the-latest-version-of-react-router-dom)

PR #855, opened 2019-02-12
Feature/helsinki brand v2

rBoost:


vikoivun:
Are you certain it should be done like this? As far I can tell, you should just change develop to point to the tip of feature/helsinki-brand-v2. It would give a more truthful view of the development.
The couple of commits that current develop has are changes specific to the old branch.

PR #857, opened 2019-02-13
[Fix] Upgrade moment, moment-range, and moment-timezone

Nizar-Rahme:


Nizar-Rahme:
[Upgrade moment, moment-timezone, moment-range](https://trello.com/c/Y6gx57ec/29-upgrade-moment-moment-timezone-moment-range)

PR #860, opened 2019-02-14
[Fix] Upgrade lodash

Nizar-Rahme:


Nizar-Rahme:
[Upgrade lodash](https://trello.com/c/XyrlsfQF/35-upgrade-lodash)

PR #862, opened 2019-02-14
[Feature] Enable redux-devtools and remove logger middleware

Nizar-Rahme:


Nizar-Rahme:
[Configure redux-devtools & Remove logger middleware](https://trello.com/c/9qeO0R2D/39-configure-redux-devtools-remove-logger-middleware)

PR #863, opened 2019-02-18
[Feature] Make homepage banner clickable

midori0507:
- Wrap div in big React-router Link component
- Use 12 col in mobile mode.

### DIFF
--- a/app/assets/styles/customization/espoo/customization.scss
+++ b/app/assets/styles/customization/espoo/customization.scss
@@ -111,11 +111,16 @@
     padding-top: 50px;
   }
 
-  .app-HomePage .app-HomePageContent__banner
+  .app-HomePage
     div {
       background-color: $blue-light;
     }
-
+  
+  .app-HomePageContent__banner {
+    &__linkWrapper {
+      background-color: $blue-light;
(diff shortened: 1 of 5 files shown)


app/assets/styles/customization/espoo/customization.scss
```diff
@@ -111,11 +111,16 @@
     padding-top: 50px;
   }
 
-  .app-HomePage .app-HomePageContent__banner
+  .app-HomePage
     div {
       background-color: $blue-light;
```
rBoost: This selector looks a bit scary to me, could we make it more precise than: "all divs under .app-HomePage"?

midori0507: agree, but i don't even know if it have some side-effect if we target only 1 div in our fix..

midori0507: Ah nvm, i guess it's just selector, will remove without any hurts

SamuelSaarentola:
[Make homepage banners clickable](https://trello.com/c/GmQJSKmy/42-make-homepage-banners-clickable)

PR #864, opened 2019-02-18
[Hotfix] admin-resources | reservation slots errors

Nizar-Rahme:


Nizar-Rahme:
[Fix admin-resources crashing errors](https://trello.com/c/1TXmJ4Pq/45-fix-admin-resources-crashing-errors)

app/shared/availability-view/_availability-view.scss
```diff
@@ -254,6 +254,9 @@
   .reservation-slot {
     height: $slot-height;
     position: relative;
+    margin: 0;
+    padding: 0;
+    border: none;
```
rBoost: Are these styles enough to make the <button> element to look like the link did before?

Nizar-Rahme: Yes, because the classNames were passed as props to the deleted component, so all the previous styles were applied to the `button` element.

rBoost: What I meant was did you make sure that the browsers default button styles don't appear?

Nizar-Rahme: Yes, at least in Chrome, let me double check in other browsers.

Nizar-Rahme: @rBoost it seems to be fine.

rBoost: 👍

PR #865, opened 2019-02-20
[Feature] Delayed Reservation

Nizar-Rahme:


Nizar-Rahme:
[Delayed reservation](https://trello.com/c/QckKEJyX/7-delayed-reservation)

Nizar-Rahme:
[Delayed reservation](https://trello.com/c/QckKEJyX/7-delayed-reservation)

app/shared/resource-calendar/ResourceCalendar.js
```diff
@@ -27,7 +27,8 @@ export class UnconnectedResourceCalendar extends Component {
     if (this.props.disableDays) {
       return this.props.disableDays(day);
     }
-    return moment(day).isBefore(moment(), 'day');
+    const today = moment().format('L');
```
rBoost: Why is this required (`.format('L')`)? Isn't the second argument of `isBefore` for defining the precision of the check?

Also maybe you could move the `const today = ...` out of the function, maybe before class definition? That way it's not run every time function is called.

Nizar-Rahme: Yes, the second argument of `isBefore` would do, no need for that `.format('L')` and that's how I tried it first, but I had a doubt that the failing unit test in travis might be because of that, apparently, it is not :( 

It's kind of impossible to put my finger on the culprit without having travis token since all unit tests are passing perfectly in my local environment.

PR #867, opened 2019-02-25
[Feature] Add `unit_has_resources` as default query when fetch units

midori0507:


app/actions/__tests__/unitActions.spec.js
```diff
@@ -0,0 +1,30 @@
...
+
+describe('Actions: unitActions', () => {
+  let getRequestTypeDescriptorMock;
+  let buildAPIUrlMock;
+
+  beforeEach(() => {
+    getRequestTypeDescriptorMock = simple.mock(apiUtils, 'getRequestTypeDescriptor');
+    buildAPIUrlMock = simple.mock(apiUtils, 'buildAPIUrl');
+  });
+
```
Nizar-Rahme: I suggest restoring mocked functions after each unit test

```suggestion

    afterEach(() => {
      simple.restore();
    });
```

midori0507: Ah, i forgot that! 👍

SamuelSaarentola:
[Remove empty slots from Unit search filter](https://trello.com/c/s09UlmIk/6-remove-empty-slots-from-unit-search-filter)

PR #868, opened 2019-02-25
[Feature] Add municipality filter, upgrade react-select

midori0507:
- Add municipality list currently as constant mock
- Add municipality selector
- Upgrade react-select to latest


app/pages/search/controls/SelectControl.js
```diff
@@ -2,36 +2,78 @@ import React, { PropTypes } from 'react';
...
-    </div>
-  );
-}
+class SelectControl extends React.Component {
+  getValue = (value, options) => {
+    if (isArray(value)) {
+      return value.map(item => options.find(option => option.value === item));
+    }
+
+    return options.find(option => option.value === value);
```
Nizar-Rahme: What if this results in `undefined`, does `react-select` handle it as a value?

app/pages/search/controls/SelectControl.js
```diff
@@ -2,36 +2,78 @@ import React, { PropTypes } from 'react';
...
-          options={options}
-          placeholder={t('common.select')}
-          searchable
-          value={value || ''}
-        />}
-      </FormGroup>
-    </div>
-  );
-}
+class SelectControl extends React.Component {
```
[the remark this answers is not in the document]
midori0507: because of getValue function, public function can not be tested

app/pages/search/controls/SelectControl.js
midori0507: Yes, i have the case in the test file, if the value is not part of options, it just get ignored

SamuelSaarentola:
[Municipaly as a filter](https://trello.com/c/K5xrt5nQ/34-municipaly-as-a-filter)

PR #869, opened 2019-02-25
[Fix] Weekly reserve calendar fixes

Nizar-Rahme:


app/pages/resource/reservation-calendar/ReservationCalendarContainer.js
```diff
@@ -80,7 +80,11 @@ export class UnconnectedReservationCalendarContainer extends Component {
     const endSlot = last(orderedSelected);
     const beginText = this.getDateTimeText(beginSlot.begin, true);
     const endText = this.getDateTimeText(endSlot.end, false);
-    return `${beginText} - ${endText}`;
+    const duration = moment.duration(moment(endSlot.end).diff(moment(beginSlot.begin)));
+    const hours = duration.hours();
+    const mins = duration.minutes();
+    const durationText = `${hours > 0 ? `${hours}h ` : ''}${mins}min`;
+    return `${beginText} - ${endText} (${durationText})`;
```
midori0507: Would it be better if we can just use moment.format with specific format here?

app/pages/resource/reservation-calendar/time-slots/TimeSlots.js
```diff
@@ -90,6 +90,9 @@ class TimeSlots extends Component {
     const firstTimeSlots = slots
       .map(timeslots => timeslots && timeslots[0])
       .filter(value => !!value && value.end);
+    const slotLength = firstTimeSlots[0]
+      ? moment(firstTimeSlots[0].end).diff(firstTimeSlots[0].start, 'm')
+      : 30;
```
midori0507: 30 should be define as constant?

app/shared/resource-calendar/ResourceCalendar.js
```diff
@@ -49,59 +54,77 @@ export class UnconnectedResourceCalendar extends Component {
...
+          <Overlay
+            container={this.calendarWrapper}
+            onHide={this.hideOverlay}
+            placement="bottom"
+            rootClose
+            show={this.state.visible}
+          >
+            <ResourceCalendarOverlay onHide={this.hideOverlay}>
+              <DayPicker
+                disabledDays={day => new Date(day).setHours(23, 59, 59, 59) < new Date()}
```
midori0507: Just feel like we should have like a local config constant in this file?

app/shared/resource-calendar/_resource-calendar.scss
```diff
@@ -2,190 +2,218 @@
...
+    }
+  
+    .DayPicker-Month {
+      margin: 0;
+      border-collapse: separate;
+    }
+  
+    .DayPicker-Day {
+      border: 2px solid $white;
+      padding: 1.2rem;
```
midori0507: IMO some  custom value like this should be store in local sass variable as much as you can

SamuelSaarentola:
[Weekly reserve calendar fixes](https://trello.com/c/iXOTB1Sg/30-weekly-reserve-calendar-fixes)

app/pages/resource/reservation-calendar/ReservationCalendarContainer.js
Nizar-Rahme: `moment.format` is not flexible enough to handle this case.

app/shared/resource-calendar/_resource-calendar.scss
Nizar-Rahme: I agree I prefer extracting all of them to `_variables` module though, which I will take care of in a new ticket.

app/shared/resource-calendar/ResourceCalendar.js
```diff
@@ -59,59 +64,77 @@ export class UnconnectedResourceCalendar extends Component {
...
+              </div>
+            </ResourceCalendarOverlay>
+          </Overlay>
+        </div>
+        <button
+          className="app-ResourceCalendar__week-button app-ResourceCalendar__week-button--next"
+          onClick={() => this.handleDateChange(
+            moment(selectedDay).add(1, 'w').toDate()
+          )
+          }
```
rBoost: Formatting.. should be `)}` instead of `)\n}`

app/shared/resource-calendar/ResourceCalendar.js
```diff
@@ -59,59 +64,77 @@ export class UnconnectedResourceCalendar extends Component {
...
-              <span className="booked">{t('ReservationCalendarPickerLegend.booked')}</span>
-            </div>
-          </ResourceCalendarOverlay>
-        </Overlay>
+        <button
+          className="app-ResourceCalendar__week-button app-ResourceCalendar__week-button--prev"
+          onClick={() => this.handleDateChange(
+            moment(selectedDay).subtract(1, 'w').toDate()
+          )
+          }
```
rBoost: Formatting.. should be `)}` instead of `)\n}`

app/shared/resource-calendar/ResourceCalendar.js
```diff
@@ -59,59 +64,77 @@ export class UnconnectedResourceCalendar extends Component {
...
+            show={this.state.visible}
+          >
+            <ResourceCalendarOverlay onHide={this.hideOverlay}>
+              <DayPicker
+                disabledDays={this.disableDays}
+                enableOutsideDays
+                initialMonth={new Date(selectedDate)}
+                locale={currentLanguage}
+                localeUtils={MomentLocaleUtils}
+                modifiers={{
```
rBoost: Could we define the modifiers outside of the markup?

app/shared/resource-calendar/ResourceCalendar.js
midori0507: @rBoost : That's what i hate about prettier..

rBoost: yep. Probably the latter one.

PR #875, opened 2019-02-28
[Feature] Expand advanced search panel when filters are applied

Nizar-Rahme:


Nizar-Rahme:
[Search: Advanced search should be open if any filters exists](https://trello.com/c/8GHLwHdb/43-search-advanced-search-should-be-open-if-any-filters-exists)

PR #876, opened 2019-02-28
[Fix] Free-of-charge filter for resources

Nizar-Rahme:


Nizar-Rahme:
[Search for only free resources does not work](https://trello.com/c/olui1sRP/10-search-for-only-free-resources-does-not-work)

PR #877, opened 2019-02-28
[Hotfix] Filter layout fix

midori0507:
- Fix `SelectControl` have different height when `has_value`
- Add better filter layout

### DIFF
--- a/app/assets/styles/_react-select.scss
+++ b/app/assets/styles/_react-select.scss
@@ -53,7 +53,8 @@ $selectDropdownPaddingRight: 4px;
     font-size: $selectFontSize;
     color: $black;
     cursor: pointer;
-
+    padding: 0;
+    
     &:hover {
       color: $black;
     }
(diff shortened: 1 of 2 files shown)


app/assets/styles/_react-select.scss
```diff
@@ -53,7 +53,8 @@ $selectDropdownPaddingRight: 4px;
     font-size: $selectFontSize;
     color: $black;
     cursor: pointer;
-
+    padding: 0;
```
rBoost: Was this causing the dropdown height to change when value was selected?

app/pages/search/controls/SearchControlsContainer.js
```diff
@@ -176,7 +176,7 @@ class UnconnectedSearchControlsContainer extends Component {
                 </Col>
               </Row>
               <Row>
-                <Col className="app-SearchControlsContainer__control" md={4} sm={6}>
+                <Col className="app-SearchControlsContainer__control" md={4} sm={12}>
```
rBoost: There is still 6 columns inside a single row, we should separate the 3 dropdowns to their own row element.

app/assets/styles/_react-select.scss
midori0507: Yes, the clear icon was 20px + 8x2 that bigger than the largest item in the select box, which push it ~2px. 

By removing this, the highest item decide the height of the select is the dropdown icon

PR #878, opened 2019-02-28
[Feature] Remove the link for old website from the footer

Nizar-Rahme:


Nizar-Rahme:
[Remove text from footer](https://trello.com/c/5faX7WYY/63-remove-text-from-footer)

PR #880, opened 2019-03-04
[Fix] Fix dateRanges return unconsistent data

midori0507:
Reservation timeslot UI only show time in current week, from start as Monday and end by Sunday. Current implementation return a day range which somehow start on Monday, end by next Monday (8 days). Sometime return 7


midori0507:
Closed by not being necessary

PR #881, opened 2019-03-04
[Hotfix] Prepend missing prefix on className, remove deprecated options

midori0507:
- Add `app-Select` as default className prefix on `SelectControl` component
- Fix deprecated prop on SelectControl on `RecurringReservation`
- Remove `classnames` import when using `SelectControl` component


Nizar-Rahme:
[Bug, repeat reservation labels missing](https://trello.com/c/726CV9Af/66-bug-repeat-reservation-labels-missing)

PR #883, opened 2019-03-05
Fix Reset Filters

Rasbin:
This PR should fix : `Reset filters` button should clear filters after it is clicked without needing to click the `Search` button after the `Reset filter` is clicked.

### DIFF
--- a/app/pages/search/controls/SearchControlsContainer.js
+++ b/app/pages/search/controls/SearchControlsContainer.js
@@ -108,6 +108,7 @@ class UnconnectedSearchControlsContainer extends Component {
     if (this.props.position) {
       this.props.actions.disableGeoposition();
     }
+
     this.handleFiltersChange(emptyFilters);
     this.handleSearch(emptyFilters);
   };
--- a/app/pages/search/controls/SearchControlsContainer.js
+++ b/app/pages/search/controls/SearchControlsContainer.js
@@ -110,6 +110,7 @@ class UnconnectedSearchControlsContainer extends Component {
     }
 
     this.handleFiltersChange(emptyFilters);
+    this.handleSearch(emptyFilters);
   };
 
   hasAdvancedFilters() {
--- a/app/pages/search/controls/SearchControlsContainer.js
+++ b/app/pages/search/controls/SearchControlsContainer.js
@@ -108,7 +108,6 @@ class UnconnectedSearchControlsContainer extends Component {
     if (this.props.position) {
       this.props.actions.disableGeoposition();
     }
-
     this.handleFiltersChange(emptyFilters);
     this.handleSearch(emptyFilters);
   };
--- a/app/pages/search/controls/SelectControl.js
+++ b/app/pages/search/controls/SelectControl.js
@@ -13,6 +13,7 @@ class SelectControl extends React.Component {
     if (isArray(value)) {
       return value.map(item => options.find(option => option.value === item));
     }
+
     return options.find(option => option.value === value);
   };
 
--- a/app/pages/search/controls/SelectControl.js
+++ b/app/pages/search/controls/SelectControl.js
@@ -13,7 +13,6 @@ class SelectControl extends React.Component {
     if (isArray(value)) {
       return value.map(item => options.find(option => option.value === item));
     }
-
     return options.find(option => option.value === value);
   };
 
--- a/app/pages/search/controls/SearchControlsContainer.spec.js
+++ b/app/pages/search/controls/SearchControlsContainer.spec.js
@@ -445,11 +445,18 @@ describe('pages/search/controls/SearchControlsContainer', () => {
       };
       instance = getWrapper(props).instance();
       instance.handleFiltersChange = simple.mock();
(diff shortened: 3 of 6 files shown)


app/pages/search/controls/SearchControlsContainer.js
```diff
@@ -62,7 +62,7 @@ class UnconnectedSearchControlsContainer extends Component {
     const { filters, position } = this.props;
     let hasFilters = Boolean(position);
     ['freeOfCharge', 'end', 'distance', 'duration', 'purpose', 'start', 'unit'].forEach((key) => {
-      if (filters[key]) {
+      if (filters[key] && filters[key] !== '0') {
```
rBoost: Why is this needed?

rBoost:
> I think you misunderstand the ticket, the implement should be like dispatch and resolve the reset action immediately after you press the button, not after you hit Search.
> Therefore, this fix is not valid, either make no sense to add the extra condition that look strange

If you check the `handleSearch` method you see that this PR does exactly that. My only question is what does the change to `hasAdvancedFilters` do, why is the change needed?

app/pages/search/controls/SearchControlsContainer.js
Rasbin: To remove the display of `Reset Filters` button after it is clicked I had to add that line. It seems like the filter value of duration has been provided zero (in number, not in string) and treated at false and affected. So, to fix that I added that code. 
screen shot 2019-03-06 at 10 04 30

rBoost: ok. I think we need to improve how search filters are put to the url. I don't think empty filters should be included in the url at all. But that's an entirely different task.

midori0507:
> > I think you misunderstand the ticket, the implement should be like dispatch and resolve the reset action immediately after you press the button, not after you hit Search.
> > Therefore, this fix is not valid, either make no sense to add the extra condition that look strange
> 
> If you check the `handleSearch` method you see that this PR does exactly that. My only question is what does the change to `hasAdvancedFilters` do, why is the change needed?

The thing is 

> > I think you misunderstand the ticket, the implement should be like dispatch and resolve the reset action immediately after you press the button, not after you hit Search.
> > Therefore, this fix is not valid, either make no sense to add the extra condition that look strange
> 
> If you check the `handleSearch` method you see that this PR does exactly that. My only question is what does the change to `hasAdvancedFilters` do, why is the change needed?

Yes you're right, i didn't notice small line change below.
Anyway, what @Rasbin trying to do here (small thing left) is the reset filter button suppose to disappear after clicked but it's not (randomly)
I suggest writing unit tests for that behavior, if the unit test passing but the issue still persist, we got problem else where, not in the code of this function

app/pages/search/controls/SearchControlsContainer.js
midori0507: We discuss and agree that this condition is not making any change (unnecessary)

app/pages/search/controls/SearchControlsContainer.js
```diff
@@ -121,6 +121,7 @@ class UnconnectedSearchControlsContainer extends Component {
     }
 
     this.handleFiltersChange(emptyFilters);
+    this.handleSearch(emptyFilters);
```
midori0507: Can you write unit test for this, something as 
`when reset button was click, trigger handleSearch`

app/pages/search/controls/SearchControlsContainer.js
Rasbin: Now, it is also working without the extra condition. It was not working before. Will investigate a bit and commit after checking if it works correct consistently.

app/pages/search/controls/SearchControlsContainer.js
Rasbin: It already failed with couple of manual testing. I will find the cause and fix the bug first.

rBoost:
@Rasbin It seems that there are some conflicts that need to be resolved before merging this issue. 
Also, we should make sure that test coverage doesn't go down.

app/pages/search/controls/SelectControl.js
```diff
@@ -13,7 +13,6 @@ class SelectControl extends React.Component {
     if (isArray(value)) {
       return value.map(item => options.find(option => option.value === item));
     }
-
```
midori0507: Can you remove this ?

app/pages/search/controls/SearchControlsContainer.js
midori0507: But still, write unit tests for it

Rasbin: I meant the reset button was not working as intended with couple of manual checks and wanted to fix it first. Missed to write it clearly before. Sorry, My bad!

app/pages/search/controls/SelectControl.js
Rasbin: Done!

Rasbin:
@rBoost Can you check it now if it is okay?

rBoost:
Looks good 👍

PR #888, opened 2019-03-08
[WIP] - Add fix for reset filter bug and update test

Rasbin:
This PR should fix : Reset filters button should clear filters after it is clicked without needing to click the Search button after the Reset filter is clicked.


Rasbin:
This is a duplicate PR and not valid anymore, already fixed in another . Could you please delete this one who have access right to delete.

PR #889, opened 2019-03-11
Disable reservation time limit for admins

Nizar-Rahme:


Nizar-Rahme:
[Maximum duration of the reservation: xx h is limited across all permission levels](https://trello.com/c/JZaFB8OH/75-maximum-duration-of-the-reservation-xx-h-is-limited-across-all-permission-levels)

Nizar-Rahme:
Currently, just checking whether the user is an admin or not `is_staff: true/false`, as other roles are not implemented.

PR #891, opened 2019-03-11
Fix reservation cancel modal UI get stuck

midori0507:
`[...state, Object.assign({})]` 
Trying to spread an Immutable object with normal object is wrong and cause the app to crash when trying to make array

Maintainance job: Upgrade `seamless-immutable` to latest, tweak README and lock node version on `package.json`

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ User interface for the City of Helsinki [varaamo.hel.fi](https://varaamo.hel.fi/
 Recommended requirements
 ------------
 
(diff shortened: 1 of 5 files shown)


midori0507:
[When canceling reservation, system gets stuck](https://trello.com/c/NF7kUBIp/76-when-canceling-reservation-system-gets-stuck)

rBoost:
Could you rename this PR and remove the [Hotfix] prefix. Lets only use [WIP] prefix.

PR #892, opened 2019-03-12
Remove unnecessary outdated dependencies

midori0507:
- Remove `react-document-title`, use `react-helmet`
- Remove `react-body-classname`, classname append can be handled by `classnames`


app/pages/PageWrapper.spec.js
```diff
@@ -19,9 +18,9 @@ describe('pages/PageWrapper', () => {
     );
   }
 
-  test('renders DocumentTitle component', () => {
-    const title = getWrapper().find(DocumentTitle);
-    expect(title.prop('title')).toBe('Test title - Varaamo');
+  test('renders Helmet title', () => {
```
midori0507: as Helmet is not easily render a component which have prop, have to use `mount` instead of `shallow` rendering.

The code below did the same thing needed to test

app/pages/AppContainer.js
```diff
@@ -60,23 +60,23 @@ export class UnconnectedAppContainer extends Component {
...
-              <Grid>
-                <Notifications />
-              </Grid>
-              {this.props.children}
-            </div>
-            <Footer />
-          </div>
-        </DocumentTitle>
-      </BodyClassName>
+      <div className={classNames('app', getCustomizationClassName())}>
```
rBoost: Are you 100% sure this is enough, because if there are any styles that depend on body class name then this most likely doesn't work?

rBoost: We just have to be 100% sure that Vantaa/Espoo site styles work like they use to work.

midori0507: @rBoost: I tested it under Espoo/Vantaa `varaamo.vantaa.fi`, remove the class from body and move it to app, everything still work after that

rBoost: ok, good. 👍

PR #893, opened 2019-03-12
Remove unnessary persisted state library, upgrade redux and dependencies

midori0507:
Remove `redux-localstorage` and `redux-localstorage-filter`. They are using just to persist locale data to localStorage. Replaced with vanilla code.

Upgrade `redux` and dependencies, some of them for example `react-redux` and `redux-form` need to upgrade again to latest because of current React version.


app/store/middleware/persistState.js
```diff
@@ -1,10 +1,20 @@
-import { compose } from 'redux';
-import persistState from 'redux-localstorage';
-import adapter from 'redux-localstorage/lib/adapters/localStorage';
-import filter from 'redux-localstorage-filter';
+export const loadPersistedLocale = () => {
+  try {
+    const locale = localStorage.getItem('userLocale');
```
midori0507: Do this file in general need to write unit tests.. ?

rBoost: That's a good question. IMO no.

midori0507: Im writing unit tests for it, better to have some even though i trust localStorage usage

PR #895, opened 2019-03-15
Sort search results

Rasbin:


app/actions/sortActions.js
```diff
@@ -0,0 +1,36 @@
...
+import schemas from 'store/middleware/Schemas';
+import {
+  buildAPIUrl,
+  getErrorTypeDescriptor,
+  getHeadersCreator,
+  getRequestTypeDescriptor,
+  getSuccessTypeDescriptor,
+} from 'utils/apiUtils';
+
+function sortResources(orderBy) {
```
rBoost: Why is sort as its own action/api call? It's only a parameter for the search api call.

Rasbin: Corrected this in the new commit. Can you verify if it is okay?

app/constants/ActionTypes.js
```diff
@@ -44,6 +44,10 @@ export default {
     SEARCH_RESULTS_GET_REQUEST: 'SEARCH_RESULTS_GET_REQUEST',
     SEARCH_RESULTS_GET_SUCCESS: 'SEARCH_RESULTS_GET_SUCCESS',
 
+    SORT_RESULTS_GET_ERROR: 'SORT_RESULTS_GET_ERROR',
+    SORT_RESULTS_GET_REQUEST: 'SORT_RESULTS_GET_REQUEST',
+    SORT_RESULTS_GET_SUCCESS: 'SORT_RESULTS_GET_SUCCESS',
```
rBoost: Are these still needed?

Rasbin: No, missed to remove before. Removed now.

PR #901, opened 2019-03-20
Navigation links for staff

midori0507:
Add new `Maintenance` tab, `Guide` tab in navbar, for admin only, one redirect to respa admin UI, one  point to gitbook.

Maintenance work: 
- Replace all `react-fontawesome` with `react-fontawesome` from `FontAwesome` author.
- Replace all `react-fontawesome` import with single component.
- Add bold theme for react-fontawesome as request from @SamuelSaarentola


midori0507:
[Navigation links for staff](https://trello.com/c/uBBGMYgL/12-navigation-links-for-staff)

PR #904, opened 2019-03-22
Favourite Resource on search view

midori0507:
PR content:
- Add ability to favourite resources straight on search view.

Extra:
- Split repeated code into reusable component.
- Merge files into 1 big container.
- Unit tests.
- Fix `jest --watch` throw out error

There are nothing to do with coverage going down, testing Redux `mapDispatchToProps` is not suggested

https://stackoverflow.com/questions/41148602/how-should-i-unit-test-mapdispatchtoprops-in-redux


midori0507:
[Add Favorite button to search results](https://trello.com/c/z79V1q7y/83-add-favorite-button-to-search-results)

app/shared/resource-card/ResourceCardContainer.js
```diff
@@ -1,6 +1,211 @@
...
+import ResourceCardInfoCell from './info';
 import resourceCardSelector from './resourceCardSelector';
-import ResourceCard from './ResourceCard';
+import {
+  favoriteResource,
+  unfavoriteResource
+} from 'actions/resourceActions';
 
-export default connect(resourceCardSelector)(ResourceCard);
+class ResourceCard extends Component {
```
rBoost: Why did you rename the file to `ResourceCardContainer`?

rBoost: I think the file should still be `ResourceCard`.

midori0507: Just to keep it consistent. Other file have unconnected component in same place with container too.

Imo there is actually no need of extra container file for just like few lines, can revert if its not make sense

midori0507: @rBoost: if its just about file naming, not content -> Yes i can rename it to `ResourceCard`

rBoost: ah yeah, I hate that naming convention. The filename shouldn't change whether or not is it connected to redux.
We should stop using that convention, remove the `Container` from the filename.

midori0507: Agree.

PR #909, opened 2019-04-03
VAR-23 | Show warning message for IE11 users

Rasbin:
This PR is for the issue #**VAR-23**
 
* Shows warning message to IE users about the page not displaying in their browser.


app/index.js
```diff
@@ -22,10 +23,47 @@ const finalState = Immutable(initialStoreState).merge([initialServerState, initi
...
-  <Provider store={store}>
-    <Router>{getRoutes()}</Router>
-  </Provider>,
-  document.getElementById('root')
-);
+if (browserName === 'IE') {
+  render(
+    <div>
+      <p style={warningStyle}>
+        Currently, Varaamo does not support IE11.
```
Nizar-Rahme: That is no longer true after adding the polyfill.

app/index.js
```diff
@@ -1,5 +1,6 @@
+import 'react-app-polyfill/ie11';
```
Nizar-Rahme: I'm quite sure that there will be some broken styles in IE, have checked it out?

Rasbin: There are some broken stuffs in IE, but in this PR the task is to only display the warning message to IE users. There is another task for fixing stuffs in IE.

app/index.js
```diff
@@ -22,10 +23,47 @@ const finalState = Immutable(initialStoreState).merge([initialServerState, initi
   deep: true,
 });
 const store = configureStore(finalState);
+const warningStyle = {
+  margin: '20px',
```
midori0507: no inline style please, add a className in sass and put styling in it

app/index.js
```diff
@@ -22,10 +23,47 @@ const finalState = Immutable(initialStoreState).merge([initialServerState, initi
...
+  border: '2px solid #faebcc',
+};
 
-render(
-  <Provider store={store}>
-    <Router>{getRoutes()}</Router>
-  </Provider>,
-  document.getElementById('root')
-);
+if (browserName === 'IE') {
```
midori0507: Write unit test

app/index.js
Rasbin: The error message `div` is getting rendered in IE browser and normally as before in other browsers only after adding the polyfill. Without polyfill, I am just getting the blank screen in IE.

PR #916, opened 2019-04-18
VAR-24 | Fix for IE11

Rasbin:
This PR is for the issue #VAR-24

- Fix messed up rendering and design stuffs in IE 11


app/pages/search/controls/_search-controls.scss
```diff
@@ -10,6 +10,15 @@
...
     padding: 30px 0;
   }
+  &__municipalityRow {
+    height: 62px;
+  }
+  &__filter-dropdown-row {
+    height: 77px;
+  }
+  &__radio-buttons-row {
+    height: 105px;
```
midori0507: This is not a fix, those random value are h@cks, you should find better way to handle the root of problem, not just trying to make it work

midori0507:
@Rasbin: To fix IE style issue, try to use specific `@media` target IE browser, and put custom style in that to fix, not trying to fix style which currently working in normal browser

https://stackoverflow.com/questions/20541306/how-to-write-a-css-hack-for-ie-11

midori0507:
Not valid fix, close for redo.

PR #918, opened 2019-04-26
SearchPage: Fix excess sort field visible on map tab

JsseL:
Search resources page has a sort results field which sorts the results
on the list tab. That sort field was visible also on map tab.

Screenshot from [varaamo.hel.fi/search](https://varaamo.hel.fi/search):
varaamo_order

Now the sort field and actually the whole results list is not rendered
at all if map tab is active.


badrange:
Hi @JsseL and thank you for the contribution! I am sorry that we respond timely. We have started a conversation on how to respond to contributions in the future. 

Since you created your pull request we have refactored the code so it cannot be merged as is. I have created a task in our internal Jira to follow this bug up: 

https://helsinkisolutionoffice.atlassian.net/browse/VAR-170

I hope to hear from you in the future!

PR #922, opened 2019-04-29
VAR-49 | Add municipality options to env config.

midori0507:
- Add ability to add custom municiple to .env
- Add safety check for non-string value and non-array.
- Add unit tests.
- Default options is central cities.


README.md
```diff
@@ -62,14 +62,27 @@ By default the running app can be found at `localhost:3000`.
...
+    CUSTOM_MUNICIPALITY_OPTIONS
     ```
 
+    Environment's variable guideline:
+
+    - `API_URL`:
+      Custom config to replace global application's api URL. Expected value is valid URL string.
+
+    - `CUSTOM_MUNICIPALITY_OPTIONS`:
+      Config for custom municipalities. Expected value is array of municipalities: `['example_city']`.
```
rBoost: I don't think anyone wants to add cities with `_` :)

midori0507: might just try Tampere :D

PR #924, opened 2019-04-30
VAR-50 | Add municipality to clear filter list.

midori0507:


app/shared/resource-map/mapSelector.js
```diff
@@ -19,7 +19,7 @@ const shouldMapFitBoundariesSelector = createSelector(
   urlSearchFiltersSelector,
   selectedUnitIdSelector,
   (filters, unitId) => (
-    Boolean(find(omit(filters, ['date', 'duration', 'end', 'page', 'start']), filter => filter !== '' && filter !== false)) || Boolean(unitId)
+    Boolean(find(omit(filters, ['date', 'duration', 'end', 'page', 'start', 'municipality']), filter => filter !== '' && filter !== false)) || Boolean(unitId)
```
Nizar-Rahme: This line might look better if it gets broken in two

midori0507: Well linter didnt complain so im just leave it be :D :D

PR #925, opened 2019-05-03
VAR-57 | Add slot_size and minPeriod to reservation select

midori0507:
- Refactor Timeslot reducer / functionalities.
- User now only able to clear the selection from clicking the clear button
- Add various unit tests / time utils.
- Populate time slot when select with `minPeriod` if resource minPeriod exist
- Add info about min period on resource page


app/pages/resource/ResourcePage.js
```diff
@@ -195,7 +195,22 @@ class UnconnectedResourcePage extends Component {
                     )}
                     {!resource.externalReservationUrl && (
                       <div>
-                        {`${t('ReservationInfo.reservationMaxLength')} ${maxPeriodText}`}
+                        {/* Show reservation max period text */}
```
rBoost: Alignments need fixing :) https://github.com/airbnb/javascript/blob/master/react/README.md#alignment

PR #927, opened 2019-05-09
VAR-51 | Translate SelectControl noOptionsMessage

jimppa:
Did not bother bothering someone for actual translations, so those are my best guess at what they could be.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ coverage
 dist
 node_modules
 npm-debug.log
+
+yarn-error\.log
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,4 @@ coverage
 dist
 node_modules
 npm-debug.log
-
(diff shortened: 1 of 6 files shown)


app/pages/search/controls/SelectControl.js
```diff
@@ -46,6 +48,7 @@ class SelectControl extends React.Component {
               isClearable={isClearable}
               isMulti={isMulti}
               isSearchable={isSearchable}
+              noOptionsMessage={this.noOptionsMessage}
```
midori0507: Need unit test for this

.gitignore
```diff
@@ -3,3 +3,5 @@ coverage
 dist
 node_modules
 npm-debug.log
+
+yarn-error\.log
```
midori0507: is the `\` typo?

PR #929, opened 2019-05-13
[WIP] refactor

midori0507:


rBoost:
We need to plan this refactoring a bit. I think we should remove everything possible from redux stores. There is absolutely no reason to store the selected time ranges, and the time that the user is hovering with his mouse. Right now Varaamo is storing a lot of data into redux store, that should be stored into a component local state.

midori0507:
> We need to plan this refactoring a bit. I think we should remove everything possible from redux stores. There is absolutely no reason to store the selected time ranges, and the time that the user is hovering with his mouse. Right now Varaamo is storing a lot of data into redux store, that should be stored into a component local state.

I will close this PR, even Redux can't help. The implementation need to be done another way.

PR #934, opened 2019-05-23
New folder structure

rBoost:
The base for the new folder structure. The idea is that now we can start gradually moving into the new folder structure.


src/routes.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
```
jimppa: Seems like this file was copied instead of moved since app/routes.js still exists, maybe delete that.

rBoost: ah, good point. I'll fix that!

rBoost: It's gone!

PR #945, opened 2019-06-06
VAR-95  | Show version number in footer

Rasbin:
https://helsinkisolutionoffice.atlassian.net/browse/VAR-95


app/shared/footer/FooterContent.js
```diff
@@ -33,6 +33,9 @@ function FooterContent({ t }) {
                 {feedbackLink}
               </p>
             </Col>
+            <Col lg={3} md={3}>
+              <p className="npm-version">v0.3.0</p>
```
midori0507: Get the value from `process.env.npm_package_version`

app/shared/footer/FooterContent.js
```diff
@@ -56,6 +59,9 @@ function FooterContent({ t }) {
...
```
midori0507: use more detail className. `app-varaamo-version` instead of generic one like `npm-version`

app/shared/footer/_footer.scss
```diff
@@ -38,4 +38,11 @@ footer {
   a:hover {
     color: #eee;
   }
+
+  .npm-version {
+    font-size: 12px;
```
midori0507: Use font-size from `opencity.design`

app/shared/footer/_footer.scss
```diff
@@ -38,4 +38,11 @@ footer {
...
+    font-family: $font-family-sans-serif;
```
midori0507: Not this. Use sass value

app/shared/footer/_footer.scss
```diff
@@ -38,4 +38,11 @@ footer {
...
+    color: #fff;
```
midori0507: Sass value

app/shared/footer/_footer.scss
```diff
@@ -38,4 +38,11 @@ footer {
...
+    line-height: 13;
```
midori0507: Why a random number here?

app/shared/footer/FooterContent.js
```diff
@@ -18,7 +19,7 @@ function FooterContent({ t }) {
     case 'ESPOO': {
       return (
         <Grid>
-          <Row>
+          <Row className="flex-container">
```
rBoost: You should never put custom classNames and styles fro Row and Column components.

app/shared/footer/FooterContent.js
```diff
@@ -33,6 +34,9 @@ function FooterContent({ t }) {
                 {feedbackLink}
               </p>
             </Col>
+            <Col className="app-varaamo-version-wrapper" lg={3} md={3}>
```
rBoost: You should never put custom classNames and styles for Row and Column components. Rather you can add a div or other wrapping element inside Column.

app/shared/footer/FooterContent.js
```diff
@@ -56,6 +60,9 @@ function FooterContent({ t }) {
...
```
rBoost: You should never put custom classNames and styles for Row and Column components. Rather you can add a div or other wrapping element inside Column.

app/shared/footer/FooterContent.js
```diff
@@ -41,7 +45,7 @@ function FooterContent({ t }) {
     case 'VANTAA': {
       return (
         <Grid>
-          <Row>
+          <Row className="flex-container">
```
rBoost: You should never put custom classNames and styles for Row and Column components. Rather you can add a div or other wrapping element inside Column.

app/shared/footer/FooterContent.js
```diff
@@ -79,6 +86,9 @@ function FooterContent({ t }) {
                 {feedbackLink}
               </p>
             </Col>
+            <Col className="app-varaamo-version-wrapper" lg={3} md={3}>
```
rBoost: You should never put custom classNames and styles for Row and Column components. Rather you can add a div or other wrapping element inside Column.

app/shared/footer/_footer.scss
```diff
@@ -6,6 +6,13 @@ footer {
   text-align: center;
   font-size: 1.6rem;
 
+  .flex-container {
```
rBoost: Why do you need this flex-container? You are basically hacking the bootstrap column behavior.
If you want the columns to go on top of each other, then you should just make the columns 100% wide when in XS width. 
You can achieve this by adding `xs` prop to the columns e.g.: `<Col lg={3} md={3} xs={12}>` .

Rasbin: It was some hack for fixing the alignment of the app version text to bottom. Now, I removed it and made some modification using media query and margin-top.

PR #957, opened 2019-07-19
Fix resource page view position when search result is clicked

Rasbin:
This PR should fix the issue VAR-58

### DIFF
--- a/app/shared/resource-card/ResourceCard.js
+++ b/app/shared/resource-card/ResourceCard.js
@@ -50,9 +50,12 @@ class ResourceCard extends Component {
 
   handleLinkClick = () => {
     window.scrollTo(0, 0);
+    const scrollTop = window.pageYOffset
+      || document.documentElement.scrollTop
+      || document.body.scrollTop;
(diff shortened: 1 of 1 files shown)


app/shared/resource-card/ResourceCard.js
```diff
@@ -49,12 +49,10 @@ class ResourceCard extends Component {
   };
 
   handleLinkClick = () => {
-    const scrollTop = window.pageYOffset
-      || document.documentElement.scrollTop
-      || document.body.scrollTop;
+    window.scrollTo(0, 0);
     const { location, history } = this.props;
     const { pathname, search } = location;
-    history.replace({ pathname, search, state: { scrollTop } });
```
Nizar-Rahme: Have you checked if this `scrollTop` method is used anywhere else?

Rasbin: It is used in SearchPage. I have now keep scrollTop as it is before and used ```window.scrollTo(0,0)``` to take the browser view to the top.

PR #960, opened 2019-08-01
VAR-134 | Add manage reservation link in page header

Rasbin:
This tiny PR adds 'Manage Reservation' link in the page header.

### DIFF
--- a/app/shared/main-navbar/MainNavbar.js
+++ b/app/shared/main-navbar/MainNavbar.js
@@ -65,7 +65,7 @@ class MainNavbar extends React.Component {
                 </NavItem>
               </LinkContainer>
             )}
-            {isLoggedIn && (
+            {isLoggedIn && isAdmin && (
               <LinkContainer to="/manage-reservations">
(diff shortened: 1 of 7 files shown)


app/shared/main-navbar/MainNavbar.js
```diff
@@ -65,6 +65,13 @@ class MainNavbar extends React.Component {
                 </NavItem>
               </LinkContainer>
             )}
+            {isLoggedIn && (
```
midori0507: have to be admin/staff also, missing isAdmin / isStaff or whatever

app/shared/main-navbar/MainNavbar.js
```diff
@@ -65,6 +65,13 @@ class MainNavbar extends React.Component {
...
+              <LinkContainer to="/manage-reservations">
+                <NavItem eventKey="manage-reservations" onClick={() => this.collapseItem()}>
+                  {t('Navbar.manageReservations')}
+                </NavItem>
+              </LinkContainer>
+            )}
```
midori0507: If you can, please do some dummy test ( can copy from navbar test, they're pretty similar )

PR #964, opened 2019-08-06
Fix resource name and details to render in swedish selection

Rasbin:


src/domain/resource/card/ResourceCard.js
```diff
@@ -68,7 +68,7 @@ class ResourceCard extends React.Component {
       intl,
       t,
     } = this.props;
-    const locale = intl.locale;
+    const locale = intl.locale === 'se' ? 'sv' : intl.locale;
```
rBoost: This is a hack, this issue should probably be fixed in `dataUtils.getLocalizedFieldValue()`.

Rasbin:
Got fixed when replacing se to sv for Swedish locale.

PR #966, opened 2019-08-06
VAR-146 | Replace "se" with "sv" for Swedish locale

Nizar-Rahme:


app/state/selectors/translationSelectors.js
```diff
@@ -8,7 +8,7 @@ import constants from '../../constants/AppConstants';
 
 function currentLanguageSelector(state) {
   const locale = state.intl ? state.intl.locale : constants.DEFAULT_LOCALE;
-  return locale === 'se' ? 'sv' : locale;
+  return locale;
```
midori0507: `return state.intl ? ...`

PR #968, opened 2019-08-07
VAR-18 Support for payments in Varaamo

jimppa:
Adds support for payments in varaamo. Includes some styling for the old reservation form. Also changes the reservation calendars price calculation to use respa instead.

Resources having a `products`-field (an array of products) should be payed for when reserving.
Some new fields required by payments have been added to the payment form (`billing_X`), those fields are managed in respa. Payment-reservations are done by adding an `order`-field to the data that is POSTed to respa for a preliminary reservation. The response gives back an URL that the user should be redirected to in order to complete the payment and finalize the reservation. When the payment is done (or interrupted) the user is redirected back to varaamo (`/reservation-payment-return`) that shows the status of the reservation.


app/pages/resource/reservation-calendar/ReservationCalendarContainer.js
```diff
@@ -57,20 +59,33 @@ export class UnconnectedReservationCalendarContainer extends Component {
     timeSlots: PropTypes.array.isRequired,
   };
 
-  state= {
+  state = {
     reservationPrice: null,
```
Nizar-Rahme: For consistency purposes with other components, this could be an empty string;

app/pages/reservation/reservation-confirmation/ReservationConfirmation.js
```diff
@@ -18,24 +21,28 @@ class ReservationConfirmation extends Component {
...
-    this.props.history.replace('/my-reservations');
+  state = {
+    reservationPrice: null,
+  }
+
+  componentDidMount() {
+    const { reservation, resource } = this.props;
+    if (hasProducts(resource)) {
+      getReservationPrice(apiClient, reservation.begin, reservation.end, resource.products)
+        .then(reservationPrice => this.setState({ reservationPrice }));
```
Nizar-Rahme: Unlike other components, `catch` is missing here.

app/pages/reservation/reservation-confirmation/_reservation-confirmation.scss
```diff
@@ -10,24 +10,31 @@
...
+      margin: 20px auto;
     }
     h3 {
       font-size: 1.6rem;
-      text-align: left;
+      font-weight: normal;
     }
   }
+  &__resource-name {
+    margin-left: 111px;
```
Nizar-Rahme: These large margin values tend to be problematic on mobile views. Have you tested it?

app/pages/reservation/reservation-phases/ReservationPhases.js
```diff
@@ -1,36 +1,49 @@
...
+  const stepMessageIds = ['information', 'confirmation'];
+
   if (isEditing) {
-    phases.splice(0, 0, 'time');
+    stepMessageIds.splice(0, 0, 'time');
   }
-  const activeIndex = indexOf(phases, currentPhase);
+
+  if (hasProducts(resource)) {
+    stepMessageIds.splice(1, 0, 'payment');
```
Nizar-Rahme: Wouldn't be better to use a logic that doesn't mutate the original array?

app/shared/reservation-date/ReservationDate.js
```diff
@@ -18,7 +18,7 @@ function ReservationDate({ beginDate, endDate }) {
   const hours = reservationEnd.diff(reservationBegin, 'hours', true);
 
   return (
-    <div className="reservation-date">
+    <div className={`${className} reservation-date`}>
```
Nizar-Rahme: And here ^

app/pages/reservation/ReservationPaymentReturnPage.js
```diff
@@ -0,0 +1,105 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import queryString from 'query-string';
+import get from 'lodash/get';
+import { camelizeKeys as camelizeKeysDeep } from 'humps';
```
midori0507: Can you leave a TODO here ? We gonna need to remove them when we refactor

app/pages/reservation/reservation-confirmation/ReservationConfirmation.js
jimppa: This is in `componentDidMount` so `state.reservationPrice` is already null here. Maybe setting an error message would be the appropriate thing to do in `catch`?

app/pages/reservation/reservation-confirmation/_reservation-confirmation.scss
jimppa: You are right, it looks bad on many different widths. Will differ a bit from layout and center this, it looks OK and avoids any hacks needed to align this text with contents from `ReservationDate`.

app/pages/reservation/reservation-phases/ReservationPhases.js
jimppa: `stepMessageIds` is declared inline (not a prop), and it's contents depends on two conditions, `isEditing` and `hasProducts(resource)`. It could be written as
```
  let stepMessageIds;
  if (isEditing && hasProducts(resource)) {
    stepMessageIds = ['time', 'information', 'payment', 'confirmation'];
  } else if (isEditing) {
    stepMessageIds = ['time', 'information', 'confirmation'];
  } else if (hasProducts(resource)) {
    stepMessageIds = ['information', 'payment', 'confirmation'];
  } else {
    stepMessageIds = ['information', 'confirmation'];
  }
```
which does make it clearer what the exact result is, but I don't think that is much better than mutating the array. Adding a new step depending on a state will also be difficult. Do you have any idea how to implement this?

Nizar-Rahme: If I understood the logic correctly, it could be implemented like: 
```javascript
let stepMessageIds = ['information', 'confirmation'];

if (isEditing) {
  stepMessageIds = [...stepMessageIds, 'time'];
}
if (hasProducts(resource)) {
  stepMessageIds = [...stepMessageIds, 'payment'];
}
```

jimppa: Ah yes, `time` would need to be first which is fine (`stepMessageIds = ['time', ...stepMessageIds]` but payment is tricker since it should be after `information` (I also now see that there is a bug if it is both editing and has products)

jimppa: So now it would be like this:
```
  let stepMessageIds = hasProducts(resource)
    ? ['information', 'payment', 'confirmation']
    : ['information', 'confirmation'];
  if (isEditing) {
    stepMessageIds = ['time', ...stepMessageIds];
  }
```

Nizar-Rahme: To me, it looks clearer 👍

app/pages/reservation/reservation-confirmation/ReservationConfirmation.js
```diff
@@ -176,7 +224,7 @@ class ReservationConfirmation extends Component {
                 t('common.addressCityLabel'),
                 reservation.billingAddressCity
               )}
-          </Well>
+          </div>
```
midori0507: Leave a TODO: clean this

PR #969, opened 2019-08-07
VAR-149 | Resource card info alway show closed status

rBoost:


app/utils/resourceUtils.js
```diff
@@ -25,10 +25,8 @@ function hasMaxReservations(resource) {
 function isOpenNow(resource) {
   const { closes, opens } = getOpeningHours(resource);
   const now = moment();
-  if (now >= moment(opens) && now <= moment(closes)) {
-    return true;
-  }
-  return false;
+
+  return now >= moment(opens) && now <= moment(closes);
```
Nizar-Rahme: How about using `moment`'s own helpers `isSameOrBefore(x)` and `isSameOrAfter(x)`?

src/domain/reservation/utils.js
```diff
@@ -0,0 +1,44 @@
...
+
+  return ongoingReservation ? moment(ongoingReservation.end) : fromMoment;
+};
+
+export const getCurrentReservation = (reservations) => {
+  const now = moment();
+  return find(reservations, (reservation) => {
+    return moment(reservation.begin) < now && now < moment(reservation.end);
+  });
+};
```
Nizar-Rahme: It would be great if we have unit tests for these utility functions.

src/domain/resource/utils.js
```diff
@@ -87,3 +95,112 @@ export const isFree = (resource) => {
...
+  if (rounded === 0) {
+    return { status: 'reserved', bsStyle: 'danger' };
+  }
+
+  return {
+    status: 'availableTime',
+    bsStyle: 'success',
+    values: { hours: rounded },
+  };
+};
```
Nizar-Rahme: And unit tests for these as well :)

src/domain/resource/utils.js
```diff
@@ -87,3 +95,112 @@ export const isFree = (resource) => {
...
+  const reservations = getOpenReservations(resource);
+
+  if (!openingHours || !openingHours.closes || !openingHours.opens) {
+    return { status: 'closed', bsStyle: 'danger' };
+  }
+
+  const nowMoment = moment();
+  const opensMoment = moment(openingHours.opens);
+  const closesMoment = moment(openingHours.closes);
+  const beginMoment = nowMoment > opensMoment ? nowMoment : opensMoment;
```
Nizar-Rahme: AFAIK calling `moment()` returns an object, therefore, using `>` for comparison won't do the trick. 
I think it's better to use `isBefore()`/`isAfter()`

rBoost: This is basically copy/pasted from the old utility functions, it's just modified to work without redux/data straight from the API.

app/utils/resourceUtils.js
rBoost: Probably yeah. This is basically copy/pasted from the old utility functions, it's just modified to work without redux/data straight from the API.

src/domain/resource/utils.js
rBoost: I didn't bother to start building tests for these, because the old code didn't have tests either.

src/domain/reservation/utils.js
rBoost: I didn't bother to start building tests for these, because the old code didn't have tests either.

PR #971, opened 2019-08-08
VAR-143 | Add fallback to available language in search page headline and desc

Rasbin:
Search page headline and description does not have fallback language currently. This PR adds fallback to available language.

### DIFF
--- a/src/common/data/utils.js
+++ b/src/common/data/utils.js
@@ -1,23 +1,16 @@
 import get from 'lodash/get';
+import values from 'lodash/values';
 
 export const getLocalizedFieldValue = (field, locale) => {
   return get(field, locale, null);
 };
 
-export const getLocalizedFieldValueWithFallback = (field, locale, fallbackLocales = []) => {
+export const getLocalizedFieldValueWithFallback = (field, locale) => {
   const localeValue = getLocalizedFieldValue(field, locale);
 
   if (localeValue) {
     return localeValue;
   }
 
-  let fallbackValue = null;
-  fallbackLocales.find((fallbackLocale) => {
-    fallbackValue = getLocalizedFieldValue(field, fallbackLocale, null);
-
-    return fallbackValue;
-  });
-
-
-  return fallbackValue;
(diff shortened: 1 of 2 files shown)


src/common/data/utils.js
```diff
@@ -3,3 +3,21 @@ import get from 'lodash/get';
...
 };
+
+export const getLocalizedFieldValueWithFallback = (field, locale, fallbackLocales = []) => {
+  const localeValue = getLocalizedFieldValue(field, locale);
+
+  if (localeValue) {
+    return localeValue;
+  }
+
+  let fallbackValue = null;
```
jimppa: You should not be making side-effects (assigning fallbackValue) in `find`, a `forEach` would be more suitable for that kind of stuff. That said, `find` is the correct method you want to be using here, but in a bit different manner. From [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find): 

> The find() method returns the value of the first element in the array that satisfies the provided testing function. Otherwise undefined is returned.

So in this case you should return `true` in the callback for `find` when the value is something that you want to use, or `false` when it's an undesired value. In this case `getLocalizedFieldValue` returns `null` when a value is not found, so in the callback for `find` you should check that whatever `getLocalizedFieldValue` returns is not null e.g. `return getLocalizedFieldValue(field, value) !== null`.

`getLocalizedFieldValue` takes two arguments, `field` and `locale` so you should not be passing any more arguments than necessary to it (remove the `null` argument).

Since `find` returns the first value that we can use, it means that you can assign the return value of `fallbackLocales.find(...)` directly to `fallbackValue`, `const fallbackValue = fallbackLocales.find(...)`

rBoost:
I wouldn't create a separate function for the fallback. I would rather just add a boolean argument for the original function.

rBoost:
> I wouldn't create a separate function for the fallback. I would rather just add a boolean argument for the original function.

`export const getLocalizedFieldValue = (field, locale, fallback = false) => {`

PR #974, opened 2019-08-09
VAR-39 | Fixed time zone

Nizar-Rahme:


app/utils/timeUtils.js
```diff
@@ -225,6 +225,26 @@ function getTimeDiff(startTime, endTime, unit, isFloat = false) {
...
 }
 
+/**
+ * getDateWithinTZ
+ * Get Date object with the default time-zone factored in
+ *
+ * @param {Moment | Date | string | undefined} value represents a specific date or undefined for the current moment
+ * @returns {Date} date object correspondent to the supplied value
+ */
+function getDateWithinTZ(value) {
```
rBoost: Why do we need this function? Why not `moment().toDate()`? https://momentjs.com/docs/#/displaying/as-javascript-date/

midori0507: Same question

Nizar-Rahme: https://github.com/moment/moment-timezone/issues/309#issuecomment-193885034

midori0507:
Close for unnecessary and will re-open when needed

PR #976, opened 2019-08-13
VAR-137 | Toggle reservation information modal for manage reservations

midori0507:


src/domain/reservation/manage/information/ManageReservationsInfoModal.js
```diff
@@ -0,0 +1,78 @@
...
+  Modal, Row, Col, Button
+} from 'react-bootstrap';
+import PropTypes from 'prop-types';
+import get from 'lodash/get';
+
+import ManageReservationsStatus from '../status/ManageReservationsStatus';
+import injectT from '../../../../../app/i18n/injectT';
+import { getDateAndTime } from '../list/ManageReservationsList';
+
+const ManageReservationsInfoModal = ({
```
rBoost: The component name and folder name doesn't match (information vs info). I think that the component should be called `ManageReservationsReservationModal` and folder should be named as `reservationModal` or `ManageReservationsReservationInfoModal` and `reservationInfoModal`.

src/domain/reservation/manage/action/ManageReservationsDropdown.js
```diff
@@ -1,27 +1,34 @@
...
         title={t('ManageReservationsList.actionsHeader')}
       >
-        <MenuItem>{t('ManageReservationsList.actionLabel.information')}</MenuItem>
+        <MenuItem onClick={() => toggleModal(true)}>{t('ManageReservationsList.actionLabel.information')}</MenuItem>
         <MenuItem>{t('ManageReservationsList.actionLabel.approve')}</MenuItem>
         <MenuItem>{t('ManageReservationsList.actionLabel.deny')}</MenuItem>
         <MenuItem>{t('ManageReservationsList.actionLabel.edit')}</MenuItem>
       </DropdownButton>
+
+      {isOpen && <ManageReservationsInfoModal isOpen={isOpen} onHide={toggleModal} reservation={reservation} />}
```
rBoost: The modal should be rendered in the `ManageReservationsPage` and opening it should be handled with callbacks. It's really weird that the modal is handled purely in the dropdown.

rBoost: We want to do the API requests in the `ManageReservationsPage`.

midori0507: The modal need selected `reservation` data when clicked, put it in `ManageReservationPage` will mean we have to have `selectedReservation` also in `ManageReservationsPage` state

rBoost: You can achieve this with callback functions. You don't need the `selectedReservation` in the actions dropdown. You shouldn't have it in there rather.

src/domain/reservation/manage/information/ManageReservationsInfoModal.js
midori0507: We have `reservation` folder already,  and this file was under it so i guess `ManageReservationsInformationModal` will fit? folder will be `informationModal`
or simply just `ManageReservationsModal` and folder name is `modal`

We wont have any other modal this in dir

rBoost: Well, it's just a matter of semantics. `ManageReservations` is the manage reservations feature/page, and that modal is a modal for showing/editing reservation information. It's not for showing "manage reservations" information.

rBoost: What does `ManageReservationsInformationModal` mean?

midori0507: Then bring ti out of `manage` folder and do a new folder under `resource` name `informationModal` ?

rBoost: Sounds better, and component name would be `ReservationInformationModal`. Or maybe just `modal/ReservationModal.js`?

midori0507: `modal` sound do better

src/domain/reservation/manage/action/ManageReservationsDropdown.js
```diff
@@ -22,6 +25,8 @@ const ManageReservationsDropdown = ({ t }) => {
 };
 
 ManageReservationsDropdown.propTypes = {
-  t: PropTypes.func
+  t: PropTypes.func,
+  toggleModal: PropTypes.func,
```
rBoost: I don't thin `toggleModal` is a good callback function name. This component doesn't know what does the callback do. It should be named into something like `onInfoClick`.

src/domain/reservation/manage/list/ManageReservationsList.js
```diff
@@ -79,6 +81,7 @@ ManageReservationsList.propTypes = {
   reservations: PropTypes.array,
   units: PropTypes.array,
   intl: intlShape,
+  toggleModal: PropTypes.func
```
rBoost: Same here, `toggleModal` is not a good name for this callback.

midori0507: Renamed

PR #978, opened 2019-08-15
VAR-152 | Show calendar data with FullCalendar

rBoost:


app/assets/styles/main.scss
```diff
@@ -10,6 +10,10 @@
 @import '~react-notifications/lib/notifications.css';
 @import '~react-toggle/style.css';
 
+@import '~@fullcalendar/core/main.css';
```
midori0507: Isnt `~` is enough ? instead of `~@`

PR #979, opened 2019-08-15
VAR-133 | Include resource detail data in reservation data

midori0507:


rBoost:
Are you sure that the resources you get from reservations through `include: 'resource_detail'` have the full unit object instead of unit id?

midori0507:
@rBoost: It does https://respa.koe.hel.ninja/v1/reservation/?page_size=50&include=resource_detail

PR #981, opened 2019-08-16
VAR-94 | Add specific terms in new reservation form

Rasbin:
`Specific terms` should be added under new reservation form as a plain text below terms checkbox and above save button.


app/pages/reservation/reservation-information/ReservationInformationForm.js
```diff
@@ -285,7 +285,7 @@ class UnconnectedReservationInformationForm extends Component {
               t('common.billingEmailAddressLabel'),
             )
           }
-          <h2 className="app-ReservationPage__title">{t('ReservationInformationForm.eventInformationTitle')}</h2>
+          <h2 className="app-renderTermsField">{t('ReservationInformationForm.eventInformationTitle')}</h2>
```
jimppa: Why did you change the className here?

Rasbin: Mistakenly got changed. Reverted back in latest commit.

PR #991, opened 2019-08-27
[Hotfix] VAR-159 | Datepicker width getting too small

midori0507:
- Almost revert old PR.
- Revert/removal overall border styling
- Give dynamic width only with recurring controls (middle aligned)

### DIFF
--- a/app/shared/date-picker/_date-picker.scss
+++ b/app/shared/date-picker/_date-picker.scss
@@ -17,10 +17,6 @@
 
   input {
     width: 100%;
-
-    &:focus {
-      outline: none
-    }
   }
 
   .date-picker-overlay {
@@ -40,10 +36,6 @@
     max-width: 400px;
     margin: 0 auto;
     width: 100%;
-
-    &:focus {
-      outline: none;
-    }
   }
 
   .DayPicker-NavBar {
@@ -92,9 +84,6 @@
     border-radius: 0;
     font-size: 1.6rem;
 
-    &:focus {
-      outline: none;
-    }
     &--available:not(.DayPicker-Day--disabled) {
       background-color: $light-gray;
       &.DayPicker-Day--selected {
--- a/app/shared/date-picker/_date-picker.scss
+++ b/app/shared/date-picker/_date-picker.scss
@@ -17,17 +17,21 @@
 
   input {
     width: 100%;
-    border: 1px solid $input-border;
-    padding: 5px;
+
+    &:focus {
+      outline: none
+    }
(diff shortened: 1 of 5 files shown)


app/shared/date-picker/_date-picker.scss
```diff
@@ -17,17 +17,21 @@
 
   input {
     width: 100%;
-    border: 1px solid $input-border;
-    padding: 5px;
+
+    &:focus {
+      outline: none
+    }
```
Nizar-Rahme: For accessibility's sake, it's better to keep the outline on. 
Even if it doesn't look pretty, we can play around with its color.

app/shared/date-picker/_date-picker.scss
```diff
@@ -87,7 +91,7 @@
     background-color: $light-gray;
     border-radius: 0;
     font-size: 1.6rem;
-  
+
     &:focus {
       outline: none;
     }
```
Nizar-Rahme: And maybe here we can bring it back?

app/shared/recurring-reservation-controls/_recurring-reservation-controls.scss
```diff
@@ -6,4 +6,23 @@
...
   }
+
+  .date-picker {
+    height: $input-height-base;
+    display: flex;
+    justify-items: center;
+    flex-direction: column;
+    border: none;
+
+    input {
```
Nizar-Rahme: Is there a way to attach a className instead for targeting the element's name?

app/shared/date-picker/_date-picker.scss
midori0507: Really ? ._. ok then

app/shared/date-picker/_date-picker.scss
midori0507: It's just linter work, i think they better with space..

app/shared/recurring-reservation-controls/_recurring-reservation-controls.scss
midori0507: Under this component there are only 1 date-picker then i dont think its necessary

app/shared/date-picker/_date-picker.scss
Nizar-Rahme: Yeah, I meant the outline was turned off within a previous PR, maybe we can bring it back :)

midori0507: Consider it as done

PR #993, opened 2019-08-28
VAR-36 | Render recurring reservation tool only to user with admin right

Rasbin:
This PR is just a minor fix to check permission level. Now, recurring reservation tool is shown only to users who have admin right.

### DIFF
--- a/app/pages/reservation/ReservationPage.js
+++ b/app/pages/reservation/ReservationPage.js
@@ -166,7 +166,6 @@ class UnconnectedReservationPage extends Component {
       recurringReservations,
       selectedReservations,
       t,
-      isAdmin,
     } = this.props;
 
     const reservationsCount = selectedReservations.length + recurringReservations.length;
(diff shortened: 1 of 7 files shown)


app/pages/reservation/ReservationPage.js
```diff
@@ -166,14 +166,15 @@ class UnconnectedReservationPage extends Component {
...
+      isAdmin,
     } = this.props;
 
     const reservationsCount = selectedReservations.length + recurringReservations.length;
     const introText = resource.needManualConfirmation
       ? t('ConfirmReservationModal.preliminaryReservationText', { reservationsCount })
       : t('ConfirmReservationModal.regularReservationText', { reservationsCount });
 
-    return (
+    return (isAdmin && (
```
midori0507: Request to move `isAdmin` check to inside of `RecurringReservationControl` so no extra check have to be performed anymore

Rasbin: We don't use `RecurringReservationControl` in many places. So, I guess it is fine.

midori0507: Fix the problem by its root, not just to make it work. Be responsible

PR #994, opened 2019-08-28
VAR-168 | Cancel button for new reservations

Nizar-Rahme:


src/domain/resource/reservationCalendar/ResourceReservationCalendar.js
```diff
@@ -294,6 +296,16 @@ class ResourceReservationCalendar extends React.Component {
     return '';
   };
 
+  onEventRender = (info) => {
+    if (info.event.id === NEW_RESERVATION) {
+      const cancelBtn = document.createElement('span');
+      cancelBtn.append('\u00D7');
```
midori0507: :D :D Nooo, use `:after` and toggle css class by state

Nizar-Rahme: You mean instead of the `span` element?
If so, won't be able to attach an event listener to it.

midori0507: make a button, display when `selected` exist, put clickHandler on that button ?
https://fullcalendar.io/docs/unselect-callback

Nizar-Rahme: And that's exactly what I've done here, the thing is you don't have much control from the react world over the `event` element (or at least to my knowledge), this `eventRender` method passes the element rendered and you can do whatever you want with it using the good old DOM manipulation methods.

midori0507: Hmm that's quite interesting, can you work on something else atm? i want to investigate this a bit

PR #1002, opened 2019-09-10
VAR-179 | Replace failure message and add a return button

Nizar-Rahme:


app/i18n/messages/sv.json
```diff
@@ -48,7 +48,8 @@
   "common.totalPriceLabel": "Total pris",
   "common.priceWithVAT": "{price}€ (inkl. moms {vat}%)",
   "payment.title": "Betalning avbruten",
-  "payment.text": "Prova igen om 15 minuter",
+  "payment.text": "Betalning avbruten.",
+  "payment.return": "Lämna tillbaka",
```
badrange: Where did you get this translation from?

Nizar-Rahme: From @SamuelSaarentola  https://helsinkisolutionoffice.atlassian.net/browse/VAR-179

badrange: @SamuelSaarentola I'm not a native Swedish speaker, so I'm not going to argue about this, I just wanted to mention that I read "lämnä tillbaka" as "return" in the same way as you'll return a book to the library. Just FYI :-)

SamuelSaarentola: @badrange @Nizar-Rahme Im not either, more correct would be just **tillbaka**

app/pages/reservation/PaymentFailed.js
```diff
@@ -7,12 +7,18 @@ import injectT from '../../i18n/injectT';
...
   t,
+  resourceId
 }) {
   return (
     <div className="app-ReservationFailedPage__contentbox">
       <h2>{t('payment.title')}</h2>
       <p>{t('payment.text')}</p>
-      <Link className="app-ReservationFailedPage__link" to="/my-reservations">
+      {resourceId && (
+        <Link className="app-ReservationFailedPage__link" to={`/resources/${resourceId}`}>
```
midori0507: Should `history.goBack()` is better (so all queries was preserved)

Nizar-Rahme: I thought about it but, the previous page is the Bambora form, and using `goBack()` prevents you from opening this URL in a new tab.

midori0507: Hah, true!

PR #1004, opened 2019-09-11
VAR-164 | Staff cannot see normal reservations

midori0507:
- Staff now able to see all reservations
- Staff which have permissions can perform reservation edit, for example changing reservation state, comments..
- Staff without permission only can open modal, see informations. Cant comment, cant edit


src/domain/reservation/manage/list/ManageReservationsList.js
```diff
@@ -68,6 +69,8 @@ const ManageReservationsList = ({
                     onEditReservation={onEditReservation}
                     onInfoClick={() => onInfoClick(reservation)}
                     reservation={normalizedReservation}
+                    userCanCancel={reservationUtils.canUserCancelReservation(reservation)}
```
Nizar-Rahme:
```suggestion
                    userCanDelete={reservationUtils.canUserCancelReservation(reservation)}
```

midori0507: i know that's dumb but `Cancel` should go along with `Cancel`..
Delete is not one of reservation state..

Nizar-Rahme: I saw the `propTypes` in [ManageReservationsDropdown.js](https://github.com/City-of-Helsinki/varaamo/pull/1004/files#diff-dbe147be8b0229268f557adf4ec40370), they don't have `userCanCancel`, instead, they have `userCanDelete`.

midori0507: Hah, will fix 👍

PR #1006, opened 2019-09-16
VAR-163 | Manage reservation filter buttons

midori0507:
1. Add `show only` filter section, to filter reservation list.
2. Add show by `can_modify` permission button
3. Add show by `favorite` button. Check user's favorite resources and filter list base on that.

MINOR: 
1. Fix 1px glitch on manage-reservaion input box.
2. Use `favorite` instead of `favourite` (US instead of UK)


src/domain/reservation/manage/page/ManageReservationsPage.js
```diff
@@ -127,20 +139,52 @@ class ManageReservationsPage extends React.Component {
...
+    const { userFavoriteResources } = this.props;
+
+    const favoriteResourceFilter = reservation => userFavoriteResources
+      && userFavoriteResources.includes(reservation.resource.id);
+
+    const canModifyFilter = reservation => reservationUtils.canUserModifyReservation(reservation);
+
+    // Both options selected
+    if (filters.length > 1) {
+      return reservations.filter(favoriteResourceFilter && canModifyFilter);
```
jimppa: If I'm reading this correctly, only `canModifyFilter` is used for filtering. Is this intended?

midori0507: yeah it is, cause the button label is the same, so that filter also reflect that.

P/s: can_modify and can_delete was one in respa (check the sourcecode)

midori0507: Instead of making 2 duplicated line, i make it as a var :P

midori0507: Fixed

PR #1008, opened 2019-09-17
VAR-163 | Add new filters to filter reset button, keep filter state when navigate.

midori0507:
- When new filters button selected, show reset filters button.
- Keep button state as internal state, so it will persist as long as user not refresh / unmount page.

Extra:
- Rename `location` to `premise` as request from PO.
- Add extra margin between buttons as request from PO.

### DIFF
--- a/src/domain/reservation/manage/filters/ManageReservationsFilters.js
+++ b/src/domain/reservation/manage/filters/ManageReservationsFilters.js
@@ -144,7 +144,7 @@ class ManageReservationsFilters extends React.Component {
                 </Col>
               </Row>
               <Row>
-                <Col md={3}>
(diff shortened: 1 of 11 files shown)


src/domain/reservation/manage/filters/ManageReservationsFilters.js
```diff
@@ -107,75 +112,73 @@ class ManageReservationsFilters extends React.Component {
       <div className="app-ManageReservationsFilters">
         <Grid>
           <Row>
```
midori0507: This file change basically moving the a big `Row` to 2 smaller `Row`. HTML change, mostly useless nothing to review here

PR #1012, opened 2019-09-17
VAR-126 - notification style changes

jimppa:
Style changes to notifications:
  - Use helsinki design system icons. Icons are exported as white since the library used for the
     notifications uses `::before` for icon placement, making svg styling hard.
  - Remove boxes border-radius, remove opacity and set color to white. Change the box-shadow-
     color to match the type of notification.

Extra: Remove unnecessary espoo customization import, it is already imported in index.js (also vantaa customization is imported there).


app/shared/notifications/_notifications.scss
```diff
@@ -1,13 +1,28 @@
 .notification-container {
+  .notification {
+    border-radius: 0;
+    color: $white;
+    opacity: 1;
+  }
   .notification-info {
     background-color: $brand-info;
+    box-shadow: 0 0 6px $brand-info;
```
midori0507: Is there a reason for 6px?

jimppa: Not really, just prettier than the default (`0 0 12px #999`) that needed to change.

PR #1015, opened 2019-09-18
Render headline and equipment with condition check

Rasbin:
What was happening : 
- Headline title was missing in resource page 
- Headline and equipment title was rendered even when there was no any content inside them.

What this PR fixes :
- Adds the headline title in the resource page
-  Makes condition check to render headline and equipment section only if they have some content. 

Test resource link for this : [https://varaamo.test.hel.ninja/resources/awfkojswli4q](https://varaamo.test.hel.ninja/resources/awfkojswli4q)

### DIFF
--- a/app/pages/resource/resource-info/ResourceInfo.js
+++ b/app/pages/resource/resource-info/ResourceInfo.js
@@ -61,7 +61,7 @@ function ResourceInfo({
         </Row>
       </ResourcePanel>
 
-      { resource.equipment.length > 0 && (<Equipment equipment={resource.equipment} />) }
+      { Array.isArray(resource.equipment) && resource.equipment.length > 0 && (<Equipment equipment={resource.equipment} />) }
     </section>
   );
 }
--- a/app/pages/resource/resource-info/ResourceInfo.js
+++ b/app/pages/resource/resource-info/ResourceInfo.js
@@ -18,9 +18,14 @@ function ResourceInfo({
 
   return (
     <section className="app-ResourceInfo">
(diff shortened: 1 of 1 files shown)


app/pages/resource/resource-info/ResourceInfo.js
```diff
@@ -18,9 +18,14 @@ function ResourceInfo({
 
   return (
     <section className="app-ResourceInfo">
-      <div className="app-ResourceInfo__description">
-        {resource.description && <WrappedText openLinksInNewTab text={resource.description} />}
-      </div>
+      {resource.description && (
+        <ResourcePanel header={t('ResourceInfo.descriptionTitle')}>
+          <div className="app-ResourceInfo__description">
+            {resource.description && <WrappedText openLinksInNewTab text={resource.description} />}
```
jimppa: Unnecessary condition here, `resource.description` is guaranteed to be something here since it's already checked for.

app/pages/resource/resource-info/ResourceInfo.js
```diff
@@ -56,7 +61,7 @@ function ResourceInfo({
         </Row>
       </ResourcePanel>
 
-      <Equipment equipment={resource.equipment} />
+      { resource.equipment.length > 0 && (<Equipment equipment={resource.equipment} />) }
```
jimppa: Is `resource.equipment` guaranteed to be an array even if there is no equipment? I would probably first check that it also is an array (e.g. `Array.isArray(resource.equipment) && resource.equipment.length > 0 ...`

PR #1022, opened 2019-09-20
VAR-197 | Move AboutPage to src

midori0507:


src/domain/about/AboutPage.js
```diff
@@ -0,0 +1,55 @@
...
+
+import AboutPartners from './AboutPartners';
+import PageWrapper from '../../../app/pages/PageWrapper';
+import FeedbackLink from '../../../app/shared/feedback-link/FeedbackLink';
+import injectT from '../../../app/i18n/injectT';
+import { getCurrentCustomization } from '../../../app/utils/customizationUtils';
+
+function AboutPage({ t }) {
+  const city = getCurrentCustomization() ? getCurrentCustomization().toLowerCase() : 'default';
+  // TODO: Remove me along with getCurrentCustomization stuff.
```
jimppa: I don't understand this todo, is there something planned that would change current implementation?

midori0507: I leave it as a reminder so i will clean the util function later. Just refactor purpose

PR #1025, opened 2019-09-24
VAR-173 | Maximum duration of reservations should not apply to staff

badrange:


src/domain/resource/utils.js
```diff
@@ -401,7 +401,7 @@ export const isTimeRangeReservable = (resource, start, end) => {
 
   // Reservation cannot be longer than the resources max period if max period is set.
   const maxPeriod = get(resource, 'max_period', null);
-  if (maxPeriod) {
+  if (!isStaff && maxPeriod) {
```
midori0507: Please add comment for this exception: 
Something as `As a staff, user have permission to ...`

src/domain/resource/utils.js
```diff
@@ -434,7 +434,11 @@ export const isTimeRangeReservable = (resource, start, end) => {
  * @param resource {object} Resource object.
  * @returns {boolean}
  */
-export const isFullCalendarEventDurationEditable = (resource) => {
+export const isFullCalendarEventDurationEditable = (resource, start, end, isStaff = false) => {
+  if (isStaff) {
+    return true;
```
midori0507: Same comment here

src/domain/resource/utils.js
```diff
@@ -383,7 +383,7 @@ export const isDateReservable = (resource, date) => {
  * @param end {Date|string} Either a Date object or date string that can be parsed as moment object.
  * @returns {boolean}
  */
-export const isTimeRangeReservable = (resource, start, end) => {
+export const isTimeRangeReservable = (resource, start, end, isStaff = false) => {
```
midori0507: Update / new unit test will be nice.

src/domain/resource/utils.js
badrange: I added the code comment to the function dockblock , that should be more correct right?

src/domain/resource/utils.js
badrange: I added the code comment to the function dockblock , that should be more correct right?

src/domain/resource/utils.js
badrange: Done!

PR #1026, opened 2019-09-24
VAR-182 | FullCalendar minPeriod

midori0507:
- Auto-select min_period time range on user time slot selection.

- If the auto-select time overlap with reserved time slot, or outside of business hours -> display error notification.

Minor:

- Refactor and bring notification message under `src`.


src/common/calendar/TimePickerCalendar.js
```diff
@@ -152,95 +116,62 @@ class TimePickerCalendar extends Component {
...
+    );
+
+    if (isValid) {
+      this.onChange({
+        start: selectionInfo.start,
+        end: minPeriodEndTime
+      });
+    } else {
+      // Display error notifications if selection is not valid
+      addNotification(selectErrorNotification(t));
```
jimppa: Do you think it would possible to just use `NotificationManager` from `react-notifications` here? It would be more in line with "let's get rid of redux stuff" that we're currently doing and it would also avoid passing down several layers of this `addNotification`-prop.

src/domain/resource/utils.js
```diff
@@ -376,6 +404,27 @@ export const isDateReservable = (resource, date) => {
   return isAdmin || (isBefore && isAfter);
 };
 
+const isBetweenReservedTimeRange = (events, start, end) => {
+  const overlapped = events.find((event) => {
```
jimppa: `Array.prototype.some` could be used instead of find, it works as find but returns a boolean.

PR #1028, opened 2019-09-24
VAR-184 | Always fetch reservations with start and end filters

badrange:
Before this change, if you go to manage reservations and change either start and end date, you'll get an empty page. The backend doesn't like queries with empty start or end dates. 

After this change, when you change start date (and end date is still empty), it automatically uses start date + 1 day as end date, and for end date, end date - 1 day for start date.


src/common/form/fields/DateField.js
```diff
@@ -51,13 +52,20 @@ class UntranslatedDateField extends React.Component {
...
       locale,
       label,
       placeholder,
       value,
     } = this.props;
     const { isOpen } = this.state;
     const date = value || moment().toDate();
+    let valueOutput;
+
+    valueOutput = value ? moment(value).format('L') : null;
```
midori0507: This is not a neat solution, this should move into a handler function instead of being as part of render

midori0507: Dumb component actually need to be dumb, those calculation should be perform where this component get used.

TLDR: Those changes should not be placed here. This is a generic component

src/domain/reservation/manage/filters/ManageReservationsFilters.js
```diff
@@ -40,6 +40,14 @@ class ManageReservationsFilters extends React.Component {
...
     } = this.props;
 
+    // Ensure that start and end are not empty. Otherwise we'll get an error from respa instead of search results.
+    if (filterName === 'start' && !filters.end) {
+      filters.end = filterValue;
+    }
+    if (filterName === 'end' && !filters.start) {
+      filters.start = filterValue;
+    }
+
```
midori0507: This check should be in DateField onChange handler, not here.
All filter changes, which mostly not related will have to go through those condition check

src/domain/reservation/manage/filters/ManageReservationsFilters.js
```diff
@@ -48,6 +56,15 @@ class ManageReservationsFilters extends React.Component {
       newFilters[filterName] = filterValue;
     }
 
+    // We want to search for events that happen until the end of the day.
```
midori0507: Take this outside of this function

src/domain/reservation/manage/filters/ManageReservationsFilters.js
```diff
@@ -136,6 +153,7 @@ class ManageReservationsFilters extends React.Component {
                 />
                 <div className="separator">-</div>
                 <DateField
+                  end
                   id="EndDateField"
                   label={t('ManageReservationsFilters.endDateLabel')}
                   onChange={(value) => {
```
midori0507: make an event handler here only for date selection change, put our logic to that

PR #1030, opened 2019-09-27
WIP: VAR-39 | Hardcode timezone for opening/business hours

badrange:
This is starting to get scary, how many other places do we need to reformat time to make things look correct?

### DIFF
--- a/src/domain/resource/utils.js
+++ b/src/domain/resource/utils.js
@@ -190,8 +190,8 @@ export const getFullCalendarBusinessHours = (resource, date = null) => {
       const dayNumber = Number(moment(item.date).format('E'));
       businessHours.push({
         daysOfWeek: [dayNumber < 7 ? dayNumber : 0],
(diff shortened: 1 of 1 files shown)


src/domain/resource/utils.js
```diff
@@ -190,8 +190,8 @@ export const getFullCalendarBusinessHours = (resource, date = null) => {
       const dayNumber = Number(moment(item.date).format('E'));
       businessHours.push({
         daysOfWeek: [dayNumber < 7 ? dayNumber : 0],
-        startTime: item.opens ? moment(item.opens).format('HH:mm') : '00:00',
-        endTime: item.closes ? moment(item.closes).format('HH:mm') : '00:00',
+        startTime: item.opens ? moment(item.opens).tz('Europe/Helsinki').format('HH:mm') : '00:00',
```
midori0507: As i debug, seems like timezone is not injected globally, we should not fix that like this

midori0507:
Close because of #1039

PR #1031, opened 2019-09-30
VAR-173 | Use resource's userPermissions.isAdmin flag to check if user isStaff

badrange:
Turns out that we nowadays should decide if a user is an admin for a resource here: 
https://respa.koe.hel.ninja/admin/resources/unitauthorization/

so we're now updating the selector

### DIFF
--- a/app/state/selectors/authSelectors.js
+++ b/app/state/selectors/authSelectors.js
@@ -24,29 +24,6 @@ function isLoggedInSelector(state) {
 }
 
 /**
- * TODO: Find out if this is needed any more, and if isn't: Remove.
- * With our current knowledge, people are given admin permissions for
- * resources through: https://respa.koe.hel.ninja/admin/resources/unitauthorization
- *
- * New staff users won't ever get any "can approve reservation" flags for any units.
- */
-const staffUnitsSelector = createSelector(
-  currentUserSelector,
-  (currentUser) => {
-    if (!currentUser.staffPerms || !currentUser.staffPerms.unit) {
-      return [];
-    }
-    const units = [];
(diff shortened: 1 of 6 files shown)


app/state/selectors/authSelectors.js
```diff
@@ -23,6 +23,13 @@ function isLoggedInSelector(state) {
   return Boolean(state.auth.userId && state.auth.token);
 }
 
+/**
+ * TODO: Find out if this is needed any more, and if isn't: Remove.
+ * With our current knowledge, people are given admin permissions for
+ * resources through: https://respa.koe.hel.ninja/admin/resources/unitauthorization
```
midori0507: I dont think we should expose this, cause varaamo is kinda open-source. This link belong to admin, which should not be opened?

app/state/selectors/authSelectors.js
```diff
@@ -40,16 +47,13 @@ const staffUnitsSelector = createSelector(
...
- * approve reservations for a resource.
+ * Check if a user has admin permission for a unit.
+ * TODO: Find a better name for this.
  */
 function createIsStaffSelector(resourceSelector) {
   return createSelector(
     resourceSelector,
-    staffUnitsSelector,
-    (resource, staffUnits) => includes(staffUnits, resource.unit)
+    resource => Boolean(resource && resource.userPermissions && resource.userPermissions.isAdmin)
```
midori0507: We should not do this for now, as discussed

midori0507:
```suggestion
    resource => Boolean(resource && resource.userPermissions && resource.userPermissions.isAdmin)
```
`get(resource, 'userPermissions.isAdmin', false)`

PR #1032, opened 2019-10-01
VAR-183 | Better max period handling for FullCalendar

midori0507:
- User able to select / resize timeslot as much as they want.

- When resized or selected time slots is smaller than min_period -> display notification, auto-select min_period amount of slots.

- When resized or selected time slots which is bigger than max_period -> display notification, auto-select amout of max_period only

- add `selectMirror` to make selection feeling same with google calendar


src/domain/resource/utils.js
```diff
@@ -436,9 +409,10 @@ const isBetweenReservedTimeRange = (events, start, end) => {
  * @param start {Date|string} Either a Date object or date string that can be parsed as moment object.
  * @param end {Date|string} Either a Date object or date string that can be parsed as moment object.
  * @param isStaff {boolean} Staff users have permission to bypass maxPeriod check.
+ * @param events {Array} Reserved reservations from resource data.
  * @returns {boolean}
  */
-export const isTimeRangeReservable = (resource, start, end, isStaff = false, events) => {
+export const isTimeRangeReservable = (resource, start, end, events) => {
```
jimppa: Was the argument `isStaff` removed by purpose? There is still code using that argument.

midori0507: Yes, this function no longer need to care about `min_period` or `max_period` anymore
Passing it is useless.

Staff member still have to respect existed reservation (no overlapping)

midori0507: Ah sheet, forgot to remove, nice find! 👍

src/domain/resource/__tests__/utils.test.js
```diff
@@ -348,36 +346,14 @@ describe('domain resource utility function', () => {
 
     expect(resourceUtils.isTimeRangeReservable(resource, `${DATE}T08:00:00Z`, `${DATE}T10:00:00Z`))
       .toBe(true);
-    expect(resourceUtils.isTimeRangeReservable(resource, `${DATE}T08:00:00Z`, `${DATE}T16:00:00Z`))
-      .toBe(false);
+
     expect(resourceUtils.isTimeRangeReservable(resource, `${DATE}T08:00:00Z`, `${DATE}T16:00:00Z`, true))
```
jimppa: I assume `events`-argument should be something else than true. Maybe test events here?

midori0507: This test for made for `max_duration` check
Apparently there are no test for `events`, that's why i remove this test

jimppa: You are right, did not notice that the test was deleted 😄 All good.

PR #1035, opened 2019-10-02
VAR-205 | Daily/ Weekly view for mobile

midori0507:
- Add custom header as well as styling only for mobile.
- Fix some broken styling on mobiles

### DIFF
--- a/src/common/calendar/TimePickerCalendar.js
+++ b/src/common/calendar/TimePickerCalendar.js
@@ -40,11 +40,24 @@ class TimePickerCalendar extends Component {
     selected: calendarUtils.getDefaultSelectedTimeRange(this.props.edittingReservation)
   };
 
-  componentDidUpdate(prevProps) {
+  componentDidMount() {
+    const { viewType } = this.state;
+
+    const calendarApi = this.calendarRef.current.getApi();
+    calendarApi.changeView(viewType);
+  }
+
+  componentDidUpdate(prevProps, prevState) {
     const { date } = this.props;
+    const { viewType } = this.state;
+
+    const calendarApi = this.calendarRef.current.getApi();
+
     if (date !== prevProps.date) {
-      const calendarApi = this.calendarRef.current.getApi();
       calendarApi.gotoDate(date);
+    }
+    if (viewType !== prevState.viewType) {
+      calendarApi.changeView(viewType);
     }
   }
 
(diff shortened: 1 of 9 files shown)


src/common/calendar/TimePickerCalendar.js
```diff
@@ -37,15 +37,33 @@ class TimePickerCalendar extends Component {
 
   state = {
     viewType: 'timeGridWeek',
-    selected: calendarUtils.getDefaultSelectedTimeRange(this.props.edittingReservation)
+    selected: calendarUtils.getDefaultSelectedTimeRange(this.props.edittingReservation),
+    header: {
+      left: 'prev,next today',
```
Nizar-Rahme: Don't really know how this really works, but I assume it should be:
```suggestion
      left: 'prev,next,today',
```

src/common/calendar/TimePickerCalendar.js
```diff
@@ -106,6 +124,26 @@ class TimePickerCalendar extends Component {
...
+  onDatesRender = (info) => {
+    const { viewType, header } = this.state;
+    let view = info.view.type;
+    let headerConfig = header;
+
+    if (window.innerWidth < 768) {
+      // mobile view config
+      view = 'timeGridDay';
+      headerConfig = {
+        left: 'today, prev',
```
Nizar-Rahme:
```suggestion
        left: 'today,prev',
```

src/common/calendar/TimePickerCalendar.js
```diff
@@ -106,6 +124,26 @@ class TimePickerCalendar extends Component {
     this.onChange(selectable);
   }
 
+  onDatesRender = (info) => {
+    const { viewType, header } = this.state;
+    let view = info.view.type;
+    let headerConfig = header;
+
+    if (window.innerWidth < 768) {
```
Nizar-Rahme: So that would apply for small screens, not necessarily mobile devices. Are we ok with that?

src/common/calendar/_timePickerCalendar.scss
```diff
@@ -33,3 +33,84 @@
   }
 }
 
+
+@media (max-width: $screen-xs-max) {
+  .fc-time-grid .fc-slats td {
```
Nizar-Rahme: Is there a way to assign custom classNames instead of targeting the existing ones?

midori0507: I think i can wrap them under `.app-Calendar` parent is the best?

src/common/calendar/TimePickerCalendar.js
midori0507: They are designed to be like that for small gap between <prev,next> and today button. Check the calendar render

src/common/calendar/TimePickerCalendar.js
midori0507: I think so, screen under 768 width is _mostly_ considered as mobile devices

PR #1044, opened 2019-10-14
[Hotfix] VAR-149 | Populate resource fetching with default time range as 1 month.

midori0507:
Otherwise reservations data under resource will be empty. This is required from respa.

Extra:

Change `passport-helsinki` dependency to install either from `git/https`. People who didnt come from Hki org can fetch and use varaamo (for open-source).

### DIFF
--- a/src/domain/reservation/manage/page/__tests__/__snapshots__/ManageReservationsPage.test.js.snap
+++ b/src/domain/reservation/manage/page/__tests__/__snapshots__/ManageReservationsPage.test.js.snap
@@ -28,7 +28,11 @@ exports[`ManageReservationsPage renders correctly 1`] = `
       </Row>
     </Grid>
     <InjectIntl(InjectT(ManageReservationsFilters))
-      filters={Object {}}
+      filters={
+        Object {
+          "date": "2019-10-14",
+        }
+      }
       onSearchChange={[Function]}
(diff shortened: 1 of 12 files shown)


src/domain/reservation/manage/page/__tests__/__snapshots__/ManageReservationsPage.test.js.snap
```diff
@@ -28,7 +28,11 @@ exports[`ManageReservationsPage renders correctly 1`] = `
       </Row>
     </Grid>
     <InjectIntl(InjectT(ManageReservationsFilters))
-      filters={Object {}}
+      filters={
+        Object {
+          "date": "2019-10-14",
```
jimppa: Is this test going to fail tomorrow?

midori0507: ah true!

midori0507: fixed

jimppa: This file still needs fixing the date to mock

PR #1059, opened 2019-11-06
VAR-224 | Cant change week when editing reservation

tuovinensanttu:
Removed black arrows from both sides of ResourceCalendar and removed tests for these buttons.  Also styled fullCalendar buttons with Helsinki colours.


src/common/calendar/_timePickerCalendar.scss
```diff
@@ -32,72 +32,76 @@
...
 
-    .fc-header-toolbar {
-      display: flex;
-
-      .fc-left,
-      .fc-right {
-        width: 30%
-      }
+    .fc-today-button {
+      border: 1px solid #000000;
```
midori0507: Use SASS variable

src/common/calendar/_timePickerCalendar.scss
```diff
@@ -32,72 +32,76 @@
...
+      border-radius: 0;
+      padding: 0.2em 0.3em;
```
midori0507: Use rem

src/common/calendar/_timePickerCalendar.scss
```diff
@@ -32,72 +32,76 @@
...
 
-        &.fc-timeGridDay-button,
-        &.fc-timeGridWeek-button {
-          &.fc-button-active {
-            display: none;
-          }
-        }
-      }
+    .fc-timeGridWeek-button,
+    .fc-timeGridDay-button {
```
midori0507: Have you check out how its look like on every screen size ?

app/shared/resource-calendar/ResourceCalendar.js
```diff
@@ -128,13 +121,6 @@ export class UnconnectedResourceCalendar extends Component {
             </ResourceCalendarOverlay>
           </Overlay>
         </div>
-        <button
```
midori0507: Any reason for removing this?
This will make the date changing not working between the datepicker and the fullcalendar

tuovinensanttu: Those  were replaced with fullCalendars next and prev buttons so changing date is still possible.

src/common/calendar/_timePickerCalendar.scss
tuovinensanttu: Replaced all to SASS variable

src/common/calendar/_timePickerCalendar.scss
tuovinensanttu: Done.

src/common/calendar/_timePickerCalendar.scss
tuovinensanttu: Added one more media query for really small screens. Looks OK.

app/shared/resource-calendar/ResourceCalendar.js
midori0507: So basically no more date-picker?

midori0507: Please remove all the styling related to this also.

And remember, there are 2 place in the app that have FullCalendar

tuovinensanttu: Datepicker still exists and works, although its date isn't updated if date is  changed via fullcalendar buttons.

Removed styling for arrows that were removed.
FullCalendar is in two places but it seems that its using the same component so styles are in both of them.

midori0507: Nice!

app/pages/reservation/reservation-time/ReservationTime.js
```diff
@@ -42,9 +43,13 @@ class ReservationTime extends Component {
       selectedReservation,
       t,
       unit,
-      isStaff
+      isStaff,
+      date
     } = this.props;
-    const date = moment(selectedReservation.begin).format('YYYY-MM-DD');
+
```
midori0507: Reservation edit page take date from editted reservation, no need for those passing, which will confuse the component

PR #1099, opened 2020-01-28
Replace build time environment with runtime environment

Jokinen:
By using a runtime environment, the app doesn't need to be rebuild on
environment changes.

This change also fixes an issue we had with the test environment. The
test environment only sets the environment variables for runtime. This
means that we don't have access to environment variables during build
time, which we assumed to be the case previously.

Ref: VAR-235


Jokinen:
I accidentally required `CUSTOM_MUNCIPALITY_OPTIONS` by directly parsing it with `JSON.parse`. I reverted this change.

app/constants/AppConstants.js
```diff
@@ -1,42 +1,48 @@
...
+
 export default {
-  API_URL: SETTINGS.API_URL,
+  API_URL: settings.API_URL,
   CUSTOMIZATIONS: {
     'varaamo.espoo.fi': 'ESPOO',
     'varaamotest-espoo.hel.ninja': 'ESPOO',
     'varaamo.vantaa.fi': 'VANTAA',
-    'varaamotest-vantaa.hel.ninja': 'VANTAA',
+    'varaamotest-vantaa.hel.ninja': 'VANTAA'
```
badrange: Is your editor automatically removing these dangling commas? I propose that you double check that this happens intentionally. Personally I prefer having comma-dangling on since diffs are easier to read. 

Something seems to be off with Varaamo's eslint settings

Jokinen: It's currently turned off in the `.eslintrc` file.

Jokinen:
I accidentally used an `es6` export mixed in with `module.exports`, fixed that now.

Jokinen:
I removed the custom `eslint` rule that disabled the `comma dangle` rules set in `airbnb`'s package. I ran `fix` which added the dangling comma to all the places it was missing from.

I cleaned the commit history on the same go.

PR #1103, opened 2020-01-31
VAR-238 | Add skip to content link

Jokinen:
This change is part of the accessibility changelist. It resolves VAR-283.

Screenshot 2020-01-31 at 9 40 01

Relevant commit in Turku's repo
https://github.com/codepointtku/varaamo/commit/2f192dea832ec9f586a8aece40c71f5e176cb942#diff-4bf6d7c1778bc725f9bbcd22e36bc8cd

### DIFF
--- a/app/i18n/messages/sv.json
+++ b/app/i18n/messages/sv.json
@@ -23,7 +23,7 @@
   "AboutPartners.aikaLogoAlt": "Logotypen för 6Aika",
   "AboutPartners.eakrLogoAlt": "Logotyp för Europeiska regionala utvecklingsfonden",
   "AboutPartners.euVipuvoimaaLogoAlt": "Hävkraft från EU-logo",
-  "SkipLink.text": "Gå direkt till innehållet",
+  "AccessibilityShortcuts.skipToContent": "Gå direkt till innehållet",
   "AdminReservationFilters.header": "Bokningens status",
   "AdminResourcesPage.noResourcesMessage": "Du har ännu inte några egna utrymmen.",
   "AdminResourcesPage.noRightsMessage": "För att visa den här sidan krävs tjänstemannabehörighet.",
--- a/app/i18n/messages/sv.json
+++ b/app/i18n/messages/sv.json
@@ -23,6 +23,7 @@
   "AboutPartners.aikaLogoAlt": "Logotypen för 6Aika",
   "AboutPartners.eakrLogoAlt": "Logotyp för Europeiska regionala utvecklingsfonden",
   "AboutPartners.euVipuvoimaaLogoAlt": "Hävkraft från EU-logo",
+  "SkipLink.text": "Gå direkt till innehållet",
   "AdminReservationFilters.header": "Bokningens status",
   "AdminResourcesPage.noResourcesMessage": "Du har ännu inte några egna utrymmen.",
   "AdminResourcesPage.noRightsMessage": "För att visa den här sidan krävs tjänstemannabehörighet.",
--- a/app/pages/AppContainer.js
+++ b/app/pages/AppContainer.js
@@ -20,6 +20,7 @@ import TestSiteMessage from '../shared/test-site-message/TestSiteMessage';
 import { getCustomizationClassName } from '../utils/customizationUtils';
 import Notifications from '../shared/notifications/NotificationsContainer';
 import UserNotificator from '../../src/common/notificator/user/UserNotificator';
+import AccessibilityShortcuts from '../shared/accessibility-shortcuts/AccessibilityShortcuts';
 
 const userIdSelector = state => state.auth.userId;
 const languageSelector = state => state.intl && state.intl.locale;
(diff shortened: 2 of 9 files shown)


app/i18n/messages/sv.json
```diff
@@ -23,6 +23,7 @@
   "AboutPartners.aikaLogoAlt": "Logotypen för 6Aika",
   "AboutPartners.eakrLogoAlt": "Logotyp för Europeiska regionala utvecklingsfonden",
   "AboutPartners.euVipuvoimaaLogoAlt": "Hävkraft från EU-logo",
+  "SkipLink.text": "Gå direkt till innehållet",
```
midori0507: This property name is different between translation files

app/pages/AppContainer.js
```diff
@@ -61,33 +62,36 @@ export class UnconnectedAppContainer extends Component {
...
   render() {
     const { isStaff, language } = this.props;
+    const mainContentId = 'main-content';
+
     return (
-      <div className={classNames('app', getCustomizationClassName())}>
-        <Helmet>
-          <title>Varaamo</title>
-        </Helmet>
+      <>
```
midori0507: This is not needed

app/pages/AppContainer.js
```diff
@@ -61,33 +62,36 @@ export class UnconnectedAppContainer extends Component {
...
+        <AccessibilityShortcuts mainContentId={mainContentId} />
```
midori0507: This belong to main content, it should wrap the main content or being on a same level with main content, not above app wrap

app/shared/accessibility-shortcuts/AccessibilityShortcuts.js
```diff
@@ -0,0 +1,26 @@
...
+
+import injectT from '../../i18n/injectT';
+
+function AccessibilityShortcuts({ t, mainContentId }) {
+  const mainContentHref = `#${mainContentId}`;
+
+  return (
+    <div className="accessibility-shortcuts">
+      <a
+        className="sr-only skip-link"
```
midori0507: Bad naming, use BEM syntax (check how className been name around app)

If this is a link use Link component from react router dom

app/shared/accessibility-shortcuts/_accessibility-shortcuts.scss
```diff
@@ -0,0 +1,18 @@
+.accessibility-shortcuts {
+  text-align: center;
+  background-color: $theme-info-light;
+  
+  .skip-link {
+    &:focus,
+    &:active {
+      position: relative;
+      left: 0;
```
midori0507: Left property on a non absolute position?

app/pages/AppContainer.js
Jokinen: This link is something we should focus first when we tab into the page from the URL bar. From that POV I think it's the safest to have it as high on the page as possible. The original audit flag suggested vero.fi (https://www.vero.fi/) as an example for implementation and they, for instance, have the shortcuts on the next level from body.

Having it within the app wrapper doesn't communicate this fact with the same "intensity".

Saying this, if you still prefer to have it within the app wrapper, I'll move it.

app/shared/accessibility-shortcuts/_accessibility-shortcuts.scss
Jokinen: I actually just copy pasted these rules. I'll try and validate that it doesn't reset anything.

app/shared/accessibility-shortcuts/AccessibilityShortcuts.js
Jokinen: Renamed these.

In this context we are not linking to another page, but just to an ID within the same page.

app/i18n/messages/sv.json
Jokinen: Good catch, fixed 👍

app/shared/accessibility-shortcuts/_accessibility-shortcuts.scss
Jokinen: So, I think this is what I've done here:  
1) I copied over the general logic for hiding elements from vero.fi
2) I copied the logic for showing elements from Turku's version of Varaamo

This created a miss match between the styles for displaying and hiding elements.

I preferred Vero.fi's approach, because they were using `sr-only`. Meaning that they are showing the content to screen readers even when it's not focused (Vero.fi is also using Bootstrap). Turku's approach may also achieve this, but the rules they were using were custom.

I've now copied the rules for "resetting" `sr-only` from vero.fi.

app/pages/AppContainer.js
midori0507: If its better to be on top, then it should be there, just my suggestion to keep things with same scope together

app/shared/accessibility-shortcuts/AccessibilityShortcuts.js
```diff
@@ -0,0 +1,26 @@
...
+import injectT from '../../i18n/injectT';
+
+function AccessibilityShortcuts({ t, mainContentId }) {
+  const mainContentHref = `#${mainContentId}`;
+
+  return (
+    <div className="app-AccessibilityShortcuts">
+      <a
+        className="sr-only app-AccessibilityShortcuts__skip-link"
+        href={mainContentHref}
```
midori0507: why dont we just `href='main-content'` here and skip all the passing props, it only used once in this case

Jokinen: So that we are not managing this ID in two places. If we pass it as a prop, we can define it in a single place. Which in this instance is the `AppWrapper`. Another route could be to have it as a constant, but I thought that this way we could have the "original" ID in the same place as before (`AppWrapper`) which sort of decreases the scope of this change.

midori0507: I just feel its not necessary to pass it around. Cause AppWrapper is the highest level and it actually wrap everything, i can't find any other usage of this

Jokinen: Here's one scenario where hard coding the value may be problematic:

If some dev changes the id in `AppWrapper`, and the dev in question is not me (:D), they likely won't know that they are breaking the functionality of the `AccessibilityShortcuts`.

My logic, in other terms, is to avoid similar sort of maintainability issues we try to avoid by using BEM (or CSS modules). By hard coding the values, we end up having implicit relationships. In this instance, because the feature in question will probably be used relatively seldomly by us, it'll be even easier to miss this relationship by accident. And in the case we would happen to miss it, it'd likely take for some time for us to spot it.

If we just hard coded the values, we could add a test that validates that the link skips to the correct place on the page. But I don't think that we currently have test cases on that level--and it'd be a more complicated way to approach this imo (although we should add a test like that if we do for instance e2e at some point).

Now that the ID is defined in a single place, the dev will, by accident, implement their changes in a way that does not break `AccessibilityShortcuts`.

I'd be willing to do this in some other way if you think it's worth it, but I would prefer to avoid having the implicit relationship here.

midori0507: Yes, approved 😄

Jokinen:
Rebased to resolve conflicts.

PR #1105, opened 2020-02-05
VAR-286 | Allow date to be inputted with keyboard in search

Jokinen:
The datepicker we use is not accessible. This was flagged in the accessibility review--the suggested fix was to enable users to input the date with a keyboard, all the while making sure that incorrect dates get accessible error labels.

As a note this fix only resolves the issue within the search page. We use another datepicker found from within `shared` components for reservations which will still behave in the same way. My thinking is that we could resolve this issue when we make the reservation flow accessible (it currently isn't because of fullcalendar).

Relevant commit/file from Turku's changes: https://github.com/codepointtku/varaamo/commit/4e2976c0d0ec8b782e84058014666ed8559c038b#diff-93e6a586300611181a31f6db860d1508


src/domain/search/filters/filter/DateFilter.js
```diff
@@ -5,21 +5,25 @@ import DayPicker from 'react-day-picker';
...
+import Button from 'react-bootstrap/lib/Button';
 
+import constants from '../../../../../app/constants/AppConstants';
 import injectT from '../../../../../app/i18n/injectT';
 import iconCalendar from './images/calendar.svg';
 
-const DatePickerWrapper = ({ children }) => (
-  <div className="app-DateFilter__datePicker">
+const DatePickerWrapper = ({ children, ...rest }) => (
+  <div aria-hidden={rest['aria-hidden']} className="app-DateFilter__datePicker" tab-index={rest['tab-index']}>
```
tuovinensanttu: In React `tab-index` should be `tabIndex`.

src/domain/search/filters/filter/DateFilter.js
```diff
@@ -48,26 +82,78 @@ class UntranslatedDateFilter extends React.Component {
...
+            />
+            <InputGroup.Button>
+              {/* We are setting tab-index as -1 in order to skip */}
+              {/* the button controlling the date picker dropdown. */}
+              {/* Because it's not keyboard accessible, it's better */}
+              {/* to help users avoid it */}
+              <Button
+                aria-hidden="true"
+                className="app-DatePickerControl__button"
+                onClick={this.handleDateButtonClick}
```
tuovinensanttu: This method doesn't exist.

src/domain/search/filters/filter/DateFilter.js
Jokinen: Good catch 👍While fixing this I noticed that there's unexpected behaviour when tabbing into this component, I'll fix that as well.

src/domain/search/filters/filter/DateFilter.js
Jokinen: Added the missing method and a test case. Kind of feel bad for letting that one through :D

Jokinen:
I ended up removing the naive attempt to block users from accessing the datepicker with tab. It's too difficult to do in practice compared to the minor benefits (if any) it provides.

Rather than blocking tabbing into the datepicker, we should just focus on not opening it for screen reader or keyboard users, which we should already be doing. If the element is not shown, the user isn't able to access it with tab.

Jokinen:
Rebased on current develop.

PR #1107, opened 2020-02-10
VAR-288 | Make language options screen reader accessible

Jokinen:
Now, when the user navigates the language menu with a screen reader, they should hear full language names instead of abbreviations. Language options should also be annotated with the correct lang attribute, enabling the screen reader to read them out using the correct language.


Jokinen:
Waiting to get release done before merging.

Jokinen:
Release will postpone, merging now.

PR #1108, opened 2020-02-10
VAR-293 | Stop screen readers from reading svg icon titles

Jokinen:
Previously some screen readers would read svg icon title's when going through the page. Oddly, that behaviour only occurred in the resource header component.

This PR pulls e7e23b4 from Turku's version of Varaamo--it sets sensible labels for the icons.

a77d173 is also pulled from Turku's version with some changes. It moves some icons from `hel-icons` into our repository in order to be able to remove hard coded titles.


badrange:
Would it make sense to change the icons upstream so that other projects won't run into the same problem?

Jokinen:
@badrange I thought about that, here's my reasoning for not doing it

1) The way in which the icons are used in Varaamo is not mentioned in [hel-icons](https://github.com/City-of-Helsinki/hel-icons) README--likely it's a case that's not officially supported or thought through
2) Last commit into [hel-icons](https://github.com/City-of-Helsinki/hel-icons) was about two years ago
3) Currently [helsinki-design-system](https://github.com/City-of-Helsinki/helsinki-design-system) seems like the more active project and it also offers icons

Due to the combination of the above conditions I think following scenarios could manifest  
1) I would need to validate the use case with someone who is responsible for [hel-icons](https://github.com/City-of-Helsinki/hel-icons). It would take time to find this person, and it would take time to validate the use case--or find a workaround.
2) Because the [hel-icons](https://github.com/City-of-Helsinki/hel-icons) project looks to be updated so seldom, and because there's a more active project that people are using for the same purpose, I doubt that there are many consumers for an updated set of icons. I.e. if I updated the upstream, I would think the interest for the new version would be limited.

Because the scope and value of this change is limited--it marginally improves the user experience in corner cases--it feels like getting these changes implemented in the upstream would be a solution that doesn't have the right balance between pragmatism and technical correctness.

I have thought about three possible strategies for resolving this issues, of which I chose this one--the one requiring the least effort   
1) Copy the commit from Turku's repo as is--trust that Varaamo's icons will be replaced with ones from HDS in time
2) Abandon `img` tag and trying and override title value by using `svg`
```
<svg>
    <title></title>
    <use xlink:href="path-to/dist/symbol/svg/hel-icons.svg#book"></use>
</svg>
```
**Problem** I couldn't reproduce the issue even though I had the same screen reader, so I am not able to validate if this fix works.

3) Pushing icons without default titles into `hel-icons`

I would personally consider doing 2 before doing 3. I am not opposed to doing the third options, especially if there are more opinions for it.

Jokinen:
Waiting to get release done before merging.

Jokinen:
Release postponed, merging now.

badrange:
@Jokinen I saw your comment only now. I wasn't aware of hel-icons being so old! I think you made a good choice by adding them here, less risk of running into a rabbit hole..

PR #1113, opened 2020-02-14
VAR-292 | Announce changes in search results and search result order

Jokinen:
Previously when you completed a search the screen reader would not give any input. The screen reader would also not give any input on changes to search result order.

Now the screen reader should read the search result count after the user searches for something (as long as it changes).

And the screen reader should also read out the current sort order after it is changed. It'll take some time for the reader to to describe the current state of the dropdown filter so it takes some patience to hear this announcement.


Jokinen:
Rebased to resolve conflict.

PR #1123, opened 2020-02-25
VAR-331, VAR-332 | Improve performance for my reservations page

Jokinen:
These changes aim to improve the performance of the my reservations page. The page takes a long time to load as of now and may even crash for devices with lower performance, for instance mobile.

## Causes for performance issues

While inspecting the performance issues I, with the help of Samuel, identified two key causes:
1) Requests were configured to fetch a very large set of data (500 pages)
2) Images were loaded in full size

### Context for large sets of data  
It looked like the reason for the large data set was that the resource object returned inlined to the reservation object is missing some fields which the list requires. To combat this, *all resources*  were fetched into the redux state so that they could be used to "hydrate" the incomplete `reservation.resource` objects. Due to the nature of redux and its normalized state, this must have been seen as a fast fix. I found a comment from some years ago that labeled the 500 page fetch as a temporary fix--into the annals with this one.

The side effect for this fetch is that it:  
1. creates a very large redux state object that harms frontend performance and
2. creates a request to respa that takes a long time to complete.

This same logic applied to units as well. But based on my understanding of the code, the `reservation.resource.unit` field actually provides all the information we need directly, so I am not entirely sure why a similar 500 page request was made for units.

## Rationale for changes

**Refactoring from redux to component state**  
In these changes I move the my-reservation container components into using component state instead of redux. This same action was completed for the manage resources page, which Samuel pointed to me as one place in the app where performance had been improved.

Migrating from redux into component state may offer some benefits when it comes to performance, but these should be marginal when compared to the lighter request we made. I still completed these changes because I trust that this direction was chosen for a valid reason before.

**Using a tabbed layout**  
Previously all the items were shown in a single list. Due to how the API works, this meant that upcoming reservations were shown in an illogical order: the reservation farthest away from this moment was first.

To resolve this issue past and upcoming reservations are now fetched separately and the user is able to change between the two by using a tab layout. I've demonstrated this change to the PO and it was accepted.

**Requesting additional resource information on demand**  
Previously, as I explained earlier, the app fetched "all the resources" into its state in order to be able to make the `reservation.resource` object complete.

After these changes the extra fields for `reservation.resource` are fetched "on demand". The conditions for "on demand" are filled when:
1) The reservation is on the page (pagination) that is currently being rendered
2) The reservation is visible in the window

This way we make about 3-4 request on first load, on smaller screens even less:
1) We fetch the reservations
2) we fetch `reservations.resource` for the first 2-3 reservations on the list

This way we:
1) Don't make that many more requests compared to before
2) We don't bloat the state
3) We don't make demanding requests to respa

As a downside I'd list:  
1) Reservation items are not completely usable on first render
     - Image will load in async manner (preferable behaviour most likely)
     - Icon alt texts are empty on first render --> should not impact accessibility that much as in most cases the `alt` attribute will have appropriate content when the user arrives to it
     - Edit button is disabled
     - Price is not shown
     - InfoLabel is not shown
     - Access code is not shown
2) We don't have "resource caching", i.e. we'll make a new fetch when we've already fetched the same resource before.

I did my best to avoid popping in the view. Currently these two things pop (based on my knowledge):  
- Access code
- Info label

**Removing additional unit fetch**  
As far I understood, there was no reason to fetch units into a separate state. To met it looked like the object in `reservation.resource.unit` was complete.

## Results

Testing the view by simply comparing its load time with Varaamo's test version should be enough to get a sense of the performance difference.

### DIFF
--- a/app/pages/user-reservations/UserReservationsPage.js
+++ b/app/pages/user-reservations/UserReservationsPage.js
@@ -15,6 +15,7 @@ import userReservationsPageSelector from './userReservationsPageSelector';
 import ReservationList from './reservation-list/ReservationListContainer';
 import * as searchUtils from '../../../src/domain/search/utils';
 import client from '../../../src/common/api/client';
+import constants from '../../constants/AppConstants';
 
 const PAGE_SIZE = 10;
 // We request past reservations from the API by using the start and end
@@ -172,7 +173,7 @@ class UnconnectedUserReservationsPage extends Component {
 
   loadPastReservations = (getFilters) => {
     const getFilterWithDefaults = (defaultFilters) => {
-      const now = moment().format('YYYY-MM-DD[T]HH:mmZZ');
+      const now = moment().format(constants.DATETIME_FORMAT);
       const filters = {
         ...defaultFilters,
         all: true,
@@ -182,7 +183,7 @@ class UnconnectedUserReservationsPage extends Component {
         // common heuristics these should not be considered to be in the
         // past, but I could not find a more efficient way to find past
         // reservations.
-        start: moment(ARBITRARY_START_DATETIME, 'YYYY-MM-DD[T]HH:mm').format('YYYY-MM-DD[T]HH:mmZZ'),
+        start: moment(ARBITRARY_START_DATETIME, 'YYYY-MM-DD[T]HH:mm').format(constants.DATETIME_FORMAT),
         end: now,
       };
 
--- a/app/pages/user-reservations/UserReservationsPage.js
+++ b/app/pages/user-reservations/UserReservationsPage.js
@@ -2,43 +2,265 @@ import PropTypes from 'prop-types';
 import React, { Component } from 'react';
 import Loader from 'react-loader';
 import { connect } from 'react-redux';
-import { bindActionCreators } from 'redux';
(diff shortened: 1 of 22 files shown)


Jokinen:
Force pushed to resolve merge conflicts.

app/pages/user-reservations/UserReservationsPage.js
```diff
@@ -2,91 +2,310 @@ import PropTypes from 'prop-types';
...
+      const filters = {
+        ...defaultFilters,
+        all: true,
+        ordering: '-begin',
+        // Note that using this method to fetch a reservation has a
+        // weakness: it'll also list currently ongoing reservations. By
+        // common heuristics these should not be considered to be in the
+        // past, but I could not find a more efficient way to find past
+        // reservations.
+        start: moment(ARBITRARY_START_DATETIME, 'YYYY-MM-DD[T]HH:mm').format('YYYY-MM-DD[T]HH:mmZZ'),
```
tuovinensanttu: `YYYY-MM-DD[T]HH:mmZZ` could be replaced with DATETIME_FORMAT constant.

Jokinen: Good suggestion👍

PR #1125, opened 2020-02-25
VAR-276 | Add support for viewer role

Jokinen:
This PR comes with a story.

When I started working on this PR, the viewer role's permissions were different to that of a manager or admin. I developed the UI in a direction that would allow for us to take singular permissions into account--because the permissions `viewer` had required this.

Previously we were naively using the user's role to allow actions within the UI--after the change we check if the role in question has the specific permissions to do an action. This is a better approach because it's maintainable and extendable, unlike using roles directly. With three roles in play it felt like a valuable change.

Because the API does not return all the permission, some of them are implicit, I replicated the permission matrix from respa into our repo. Unlike with `reservations`, we aren't able to trust the `userPermissions` object when it comes to `resources`. That forced me to create more scaffolding in our end.

Well, when I mentioned these discrepancies in the permissions, the PO found them to be incorrect. Respa was updated and the `viewer` role received more permissions. This rendered some of the changes I made obsolete.

As I tested my implementation I found more odd behaviour. I got a lot of help from the respa team in uncovering the core reasons. I was able to nail down two niche cases in detail: 

>The unit viewer role doesn't have permissions to create, edit or comment reservations which have been made for resources that don't have reservations enabled (reservable = `false`).

> Please check Slack for the other case
>https://helsinkicity.slack.com/archives/C51R13GRZ/p1582108269166700

These cases offer their own UX conundrums and do not have an apparent best answer. Because these cases are niche, we decided not to address them. Instead, we will just elevate the `viewer` role to be on par with `admin` and `manager` within the Varaamo UI. Most of the time respa will be able to send a permission error Varaamo knows how to display to the user. If niche cases raise a lot of user feedback, we'll address them.

What does this mean? The user permission system I created is now in essence useless. I've received explicit permission from the PO to ignore permission differences between the roles if it expedites our development effort. So the smallest change we require now is changing the `createIsStaffSelector` to respect the unit `viewer` role. I've done that now. 

I still have the permission specific checks in the code I managed to develop before this decision was made.

Those checks are no longer vital code, but they are not dead code either. I don't really know whether I should purge them or not.

**Pros**  
- It's already developed
- It's the direction we should go towards
- It's not dead code

**Cons**  
- It makes the permission system of Varaamo more complicated to comprehend
- Respa's resource permissions aren't completely finalised yet, the model would likely change if there was enough resources put into the project (unlikely). At some point the API will likely return the complete permissions.

So when reviewing this PR, I'd appreciate opinions on whether I should discard
* Add support for unit viewer role and
* Sync permission with 0.9.0 version of respa

commits or not.

### DIFF
--- a/app/shared/modals/reservation-info/ReservationEditForm.js
+++ b/app/shared/modals/reservation-info/ReservationEditForm.js
@@ -10,7 +10,10 @@ import FormControl from 'react-bootstrap/lib/FormControl';
 import FormGroup from 'react-bootstrap/lib/FormGroup';
 import Well from 'react-bootstrap/lib/Well';
 import { Field, Fields, reduxForm } from 'redux-form';
+import moment from 'moment';
 
+import { resourceRoles, resourcePermissionTypes } from '../../../../src/domain/resource/permissions/constants';
+import { hasPermissionForResource } from '../../../../src/domain/resource/permissions/utils';
 import { getReservationPrice, getTaxPercentage } from '../../../../src/domain/resource/utils';
 import FormTypes from '../../../constants/FormTypes';
 import ReduxFormField from '../../form-fields/ReduxFormField';
@@ -97,9 +100,36 @@ class UnconnectedReservationEditForm extends Component {
 
   renderReservationTime() {
     const {
-      isEditing, reservation, resource, t,
+      isEditing, reservation, resource, t, userUnitRole,
     } = this.props;
+
     if (isEditing) {
+      const canIgnoreOpeningHours = hasPermissionForResource(
+        userUnitRole,
+        resourcePermissionTypes.CAN_IGNORE_OPENING_HOURS,
+      );
+      const reservationDate = moment(reservation.begin).format('YYYY-MM-DD');
+      // We have a generic utility function with the same purpose, but
+      // for some reason the fields on the resource object are in
+      // camelCase when we access them here. The generic utility
+      // function assumes snake_case.
+      const getOpenHours = (resource0, reservationDate0) => {
+        const openHours = resource0.openingHours || [];
+
+        return openHours.find(openHour => openHour.date === reservationDate0);
+      };
+      const openingHoursForReservationDate = getOpenHours(resource, reservationDate);
+      const getConstraints = (openingHours) => {
+        if (!openingHours || canIgnoreOpeningHours) {
+          return undefined;
+        }
+
+        return {
+          startTime: moment(openingHours.opens).format('HH:mm'),
+          endTime: moment(openingHours.closes).format('HH:mm'),
+        };
+      };
+
       return (
         <FormGroup id="reservation-time">
           <Col sm={3}>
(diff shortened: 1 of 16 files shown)


app/shared/modals/reservation-info/ReservationEditForm.js
```diff
@@ -97,9 +100,36 @@ class UnconnectedReservationEditForm extends Component {
...
+      const canIgnoreOpeningHours = hasPermissionForResource(
+        userUnitRole,
+        resourcePermissionTypes.CAN_IGNORE_OPENING_HOURS,
+      );
+      const reservationDate = moment(reservation.begin).format('YYYY-MM-DD');
+      // We have a generic utility function with the same purpose, but
+      // for some reason the fields on the resource object are in
+      // camelCase when we access them here. The generic utility
+      // function assumes snake_case.
+      const getOpenHours = (resource0, reservationDate0) => {
```
midori0507: the argument with number is kinda bad, how about `getReservationOpeningHoursByDate` ? 😄 
Bring that to method level (same level with `renderReservationTime` or make it generic and bring it to `https://github.com/City-of-Helsinki/varaamo/blob/develop/src/domain/reservation/utils.js` instead of nesting it here

app/shared/modals/reservation-info/ReservationEditForm.js
```diff
@@ -97,9 +100,36 @@ class UnconnectedReservationEditForm extends Component {
...
-      isEditing, reservation, resource, t,
+      isEditing, reservation, resource, t, userUnitRole,
     } = this.props;
+
     if (isEditing) {
+      const canIgnoreOpeningHours = hasPermissionForResource(
+        userUnitRole,
+        resourcePermissionTypes.CAN_IGNORE_OPENING_HOURS,
+      );
+      const reservationDate = moment(reservation.begin).format('YYYY-MM-DD');
```
midori0507: replace YYY-MM-DD with DATE_FORMAT constant

src/domain/resource/permissions/utils.js
```diff
@@ -0,0 +1,67 @@
...
+
+  return null;
+}
+
+const permissionOptions = Object.values(resourcePermissionTypes);
+const allValuesInArray = (subset, superset) => difference(subset, superset).length === 0;
+
+export function hasPermissionForResource(resourceRole, requiredPermissions) {
+  // undefined roles don't receive any permissions implicitly
+  if (resourceRole === undefined || resourceRole === null) {
```
midori0507: if(!resourceRole)

midori0507:
Pretty great PR i would say. Well done.

Could be nice if you can wrote a bit documentation about your permission knowledges to some docs. (or on your permissions util file) / (Varaamo gitbook), a for-developer version.

badrange:
Great that you added documentation! This is going to be very helpful to us in the future. 

If you find the time I think it could be helpful to write down some ideas on how this could be improved in the future if there is time to reduce technical debt. 

Mentioning that the respa api and permission system has been changing drastically over time might also help developers understand why things are strange here and there .. :-)

PR #1126, opened 2020-02-27
VAR-333, VAR-334, VAR-335 | Refactor reservation control buttons in my reservations

Jokinen:
This PR contains changes that were needed due to refactors in previous commits. It also contains the changes described in VAR-333, VAR-334 and VAR-335.

### VAR-333 | Always display info button

Previously the info button was only displayed for a subset of role and permission combinations. Now it should be displayed to everyone.

### VAR-334 | Send user to edit view instead of inline modal
Previously, when the user clicked on the edit button on `ReservationInfoModal`, an inline form would open within the modal. After these changes, the button will redirect the user into the reservation edit flow instead.

### VAR-335 | Ensure that a regular user does not see the commenting field in the `ReservationInfoModal`
This required no changes--this was already the case before.

### Allow owner or resource to access edit from resource info modal
I didn't create a subtask for this item, but it was discussed with the PO. A user with no roles, or the role os "reserver", should also be able to access the edit button within the `ReservationInfoModal`.

----

Other changes and fixes

**Fix incorrect isStaff status in ReservationControls**  
I tried to find out how this had worked previously. I didn't understand how `staffUnits` was passed to the list item component previously. With a search, I only found "staffUnits" from within the the reservation litem item component file.

Regardless, I refactored this logic to rely on the `resource` object. The permission is now checked deeper within the component hierarchy which isn't necessarily a positive, but at least it's easy for future devs to see how the logic works. I refactored the permission logic as well--this stuff was mostly implementation detail. I wanted to avoid creating utilities with similar purposes.

**Fix edit button being disabled when resource is found**  
This was a regression brought on by me in the last set of changes. The boolean logic here was the opposite to what it should have been.

**Avoid refetching hydrated resource**  
This is a small optimization. Previously resources would be fetched whenever they entered the view--even if they had been fetched once already.

**Fix reservation edit button leading nowhere**   
In the last batch of changes, the data model changed in a way that I did not expect. This resulted in the edit button creating a broken link. The reservation object may have a resource that's a string id referencing a resource, or then it may have an inlined resource object that contains a subset of the fields of the resource object. Previously the inlined case was not supported.

**Fix UI crashing on second step of reservation editing**  
An issue that has the same root cause then the last one: inlined resource object in the reservation object.

~~NOTE: I'll reconsider the two fixes above change after I've implemented all the changes requested in VAR-267. I think that it may be prudent to clean the inlined resource out of the reservation before sending it into the redux state instead of supporting both versions.~~
Done in "Clean reservation before sending into redux"

**Fix ReservationInfoForm not displaying all fields**  
This was an issue that had its roots in the combined use of `snake_case` and `camelCase` within the application. `ReservationInfoModal` assumes `camelCase`, but it received an object that used `snake_case`.

**Fix reservation state changes not giving feedback**  
This component was refactored to use its internal state for tracking
reservations. However, some components lower in the hierarchy operate
within the redux state.

For instance, a reservation may be canceled. In this instance, the user
action of cancelling the reservation changes the state of the
reservation within the redux state. There're no easy ways to hook into
the actions themselves.

With this change we are trying to reflect changes within the redux
state in the list of reservations managed within the user reservations
page. This way we are able to integrate these two states. Otherwise we
would need to create new versions of the components that are lower in
the hierarchy that allow hooking up to user actions.

### DIFF
--- a/app/shared/modals/reservation-info/ReservationInfoModal.js
+++ b/app/shared/modals/reservation-info/ReservationInfoModal.js
@@ -16,6 +16,7 @@ import InfoLabel from '../../../../src/common/label/InfoLabel';
 import { isStaffEvent, getEditReservationUrl } from '../../../utils/reservationUtils';
 import ReservationEditForm from './ReservationEditForm';
 import { hasProducts } from '../../../utils/resourceUtils';
+import { RESERVATION_STATE } from '../../../../src/constants/ReservationState';
 
 class ReservationInfoModal extends Component {
   constructor(props) {
@@ -79,8 +80,8 @@ class ReservationInfoModal extends Component {
     );
     const canModify = hasPermissionForResource(userUnitRole, resourcePermissionTypes.CAN_MODIFY_RESERVATIONS);
     const showCancelButton = reservationIsEditable && (
-      reservation.state === 'confirmed'
-      || (reservation.state === 'requested' && !isAdmin)
+      reservation.state === RESERVATION_STATE.CONFIRMED
+      || (reservation.state === RESERVATION_STATE.REQUESTED && !isAdmin)
     );
 
     const paymentLabel = constants.RESERVATION_PAYMENT_LABELS[reservation.state];
--- a/app/shared/modals/reservation-info/ReservationInfoModal.js
+++ b/app/shared/modals/reservation-info/ReservationInfoModal.js
@@ -13,7 +13,7 @@ import { hasPermissionForResource } from '../../../../src/domain/resource/permis
 import injectT from '../../../i18n/injectT';
 import ReservationCancelModal from '../reservation-cancel/ReservationCancelModalContainer';
 import InfoLabel from '../../../../src/common/label/InfoLabel';
-import { isStaffEvent } from '../../../utils/reservationUtils';
+import { isStaffEvent, getEditReservationUrl } from '../../../utils/reservationUtils';
 import ReservationEditForm from './ReservationEditForm';
 import { hasProducts } from '../../../utils/resourceUtils';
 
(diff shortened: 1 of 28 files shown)


app/shared/modals/reservation-info/ReservationInfoModal.js
```diff
@@ -57,16 +72,16 @@ class ReservationInfoModal extends Component {
...
     const canComment = hasPermissionForResource(userUnitRole, resourcePermissionTypes.CAN_COMMENT_RESERVATIONS);
     const canViewComment = hasPermissionForResource(
       userUnitRole,
       resourcePermissionTypes.CAN_ACCESS_RESERVATION_COMMENTS,
     );
     const canModify = hasPermissionForResource(userUnitRole, resourcePermissionTypes.CAN_MODIFY_RESERVATIONS);
+    const showCancelButton = reservationIsEditable && (
+      reservation.state === 'confirmed'
+      || (reservation.state === 'requested' && !isAdmin)
+    );
```
tuovinensanttu: Comparisons could be replaced with `RESERVATION_STATE` constants.

app/shared/reservation-controls/ReservationControls.js
```diff
@@ -63,6 +64,18 @@ class ReservationControls extends Component {
...
+  get isStaff() {
+    const { resource } = this.props;
+
+    if (!resource) {
+      return false;
+    }
+
+    const isUnitStaff = getIsUnitStaff(getUnitRoleFromResource(resource));
+
+    return isUnitStaff;
```
tuovinensanttu:
```suggestion
   return getIsUnitStaff(getUnitRoleFromResource(resource));
```
No need to store value to variable.

PR #1128, opened 2020-02-27
VAR-266 | Usability improvements into reservation management page

Jokinen:
After these changes, approve and deny buttons should be hidden when the `reservation`'s state is not `requested` in these cases:  
1) Dropdown menu in the list of reservations
2) Information modal that the user can access by opening the dropdown and selecting "Information"

The English version of the application should also now use "Approve" instead of "Confirm" when it comes to the forementioned buttons and locations.

### DIFF
--- a/src/domain/reservation/manage/action/__tests__/ManageReservationsDropdown.test.js
+++ b/src/domain/reservation/manage/action/__tests__/ManageReservationsDropdown.test.js
@@ -5,11 +5,48 @@ import { UntranslatedManageReservationsDropdown as ManageReservationsDropdown } 
 import { shallowWithIntl } from '../../../../../../app/utils/testUtils';
 import reservation from '../../../../../common/data/fixtures/reservation';
 
+const findButtonByLabel = (wrapper, label) => wrapper.find({ children: `ManageReservationsList.actionLabel.${label}` });
+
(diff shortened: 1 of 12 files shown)


src/domain/reservation/manage/action/__tests__/ManageReservationsDropdown.test.js
```diff
@@ -5,11 +5,48 @@ import { UntranslatedManageReservationsDropdown as ManageReservationsDropdown }
...
-    expect(toJSON(wrapper)).toMatchSnapshot();
+    expect(toJSON(getWrapper())).toMatchSnapshot();
+  });
+
+  test('show information, edit and cancel buttons', () => {
+    const wrapper = getWrapper();
+
+    expect(findButtonByLabel(wrapper, 'information').length).toEqual(1);
+    expect(findButtonByLabel(wrapper, 'edit').length).toEqual(1);
+  });
```
tuovinensanttu: Based on test description this is missing check for cancel button. Adding a test where `userCanCancel` is set to `true` would be benefitical, because ManageReservations is a resource for admins after all.

Jokinen: Good catch. I fixed the description and added a test case for the cancel button 👍

PR #1134, opened 2020-03-19
VAR-340 | Fix calendar failing to render opening time of 23

Jokinen:
Previously the getFullCalendarMaxTime function would return 00:00:00
when a resources closing time was 23:00:00. This caused the max time to
be smaller than the end time, which in turn caused the calendar to not
render any timeslots.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
   - Fixed missing results in search results
   - Fixed scrollbar flickering in reservation calendar when business hours were of short duration
   - Fixed premise dropdown returning empty labels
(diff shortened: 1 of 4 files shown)


tuovinensanttu:
I tried the apply max-height as you did in #1135. This will shorten the jumping distance but not prevent it.

Jokinen:
Ok, good catch 👍 I'll try and resolve this .

Jokinen:
I added a fixed margin to the timegrid container. This makes small overflows and small changes in overflows invisible to the user. Because the event can overflow to a small degree, scroll bars won't be displayed at all.

PR #1138, opened 2020-03-24
VAR-356 | Allow uneven minute values

Jokinen:
**Trust API when configuring slot size**  
>Some resources have slot sizes that span a greater portion of the day.
For instance, a resource may have a configuration as follows: opening
time of 09:00, closing time of 15:00 and a mininum slot duration of 3
hours. In this case the day has two possible slots: 09:00-11:59 and
12:00-14:59.

>With our previous approach, we did some scaffolding to ensure that the
00:30:00 and 01:00:00 slot sizes worked well even if the API returned
with an unexpected value for slot_size. Because this would have been
laboursome to do for a more open eded set of possible slot sizes, I
removed the special checks. Now the slot size that is returned from the
API gets used as is.

>I also removed the buffer setting and use the slot size as the buffer
instead. This ensures that slots get spaced appropriately in the
calendar. The buffer setting was not in use in application code.

This change makes fullcalendar's slot size dynamic. Currently it's either `00:30:00` or `01:00:00`, after this change it can be anything.

I changed the styling so that slots are rendered taller when their slot size is bigger.

**Allow uneven minute values**  
>Some spaces are open through the entire day. In these cases the opening
hours are set to end at 23:59. If we ignore minutes, or round the
minutes up, we end up with "lost slots". In the first case we lose the
slot from 23:00-23:30, in the second case we lose all the slots.

>When we are using a closing time of 23:59, the "last" slot
(23:30-23:59) can't be reserved, but it is rendered in the calendar.

This is a fix to VAR-340. Currently it's possible to reserve times until 23:00. Now the calendar also supports reservations until 23:30. The last slot of the day is not reservable and according to the current spec it won't be made available.

### DIFF
--- a/src/domain/resource/utils.js
+++ b/src/domain/resource/utils.js
@@ -249,7 +249,7 @@ export const getFullCalendarBusinessHoursForDate = (resource, date) => {
  * @param buffer {number} buffer in hours.
  * @returns {string}
  */
-export const getFullCalendarMinTime = (resource, date, viewType, buffer = 1) => {
+export const getFullCalendarMinTime = (resource, date, viewType) => {
   const defaultMin = '07:00:00';
   let openingHours = null;
   switch (viewType) {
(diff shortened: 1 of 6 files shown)


src/domain/resource/utils.js
```diff
@@ -651,3 +610,29 @@ export const getTaxPercentage = (resource) => {
 
   return '';
 };
+
+export const getSlotTallness = (resource) => {
+  const [hours, minutes, seconds] = get(resource, 'slot_size', '00:03:00').split(':').map(val => Number(val));
```
tuovinensanttu: Should the default value be `00:30:00` instead of `00:03:00`? The method returns correct value nevertheless.

Jokinen: You are correct 👍

PR #1144, opened 2020-05-28
VAR-379 | Update refund policy

Jokinen:
Unfortunately neither can be tested, because paid reservations can't be made outside of the prod environment.

----
**Update refund policy text**
Unfortunately the version of react-intl that's in use lags behind by
two major versions. This means that there's no better way to apply
these translations. Updating react-intl would require major testing to
ensure that current views do not break.

Only Finnish translation is updated for now.

**Add refund policy into ReservationInformationModal**
This component lacked access to the complete version of the resource
its reservation is for. Unfortunately this is needed for calculating
the price of a reservation. Because of this I needed to add quite a lot
of scaffolding in order to be able to know when the refund policy
should be shown. Moreover it'll just pop into the view because we can't
know whether it should be shown before we receive data from the server.


Jokinen:
The older version of `react-intl` actually has `FormattedHTMLMessage` which would have made for a nicer translation.

PR #1146, opened 2020-05-28
VAR-382 | Show billing address when reservation is billable

Jokinen:
When a reservation is billable and the billing email address is set,
respa will send the reminder email into this address.


Jokinen:
Accidentally pushed a commit into the wrong branch. Removed it now.

PR #1151, opened 2020-06-02
VAR-386 | Fix fixed prices always being zero

Jokinen:
The utility that's used for calculating a reservation price lacked
support for fixed product prices.


Jokinen:
Resolve conflict in changelog.

PR #1153, opened 2020-06-08
VAR-358 | Use a single link in resource card title area

Jokinen:
Generally, having two links will make it harder for a screen reader
user to know where they are going. More specifically, the first link
was broken and drew away user attention from the most important click
target that will navigate the user into the resource page.


Jokinen:
I've toggled off my change where I increased the font size. I noticed, after these changes, that HDS was having an effect on heading font sizes, which guided me to adjust this font size. I'll resolve the issue with HDS in another PR.

PR #1154, opened 2020-06-09
VAR-360 | Add font change controls

Jokinen:
In this PR I add a font size control to the top navbar. It's used to change the base font size of the page and is part of accessibility requirements.


Jokinen:
Rebased to resolve merge conflicts.

@tuovinensanttu thanks for looking it through in so much detail! I checked out the problem and it looks like it's because the input is custom made by us and doesn't use bootstrap values to set its size. I'm a little pressed for time, so I'll come back to this quirk if I have time.

PR #1155, opened 2020-06-09
VAR-357 | Add keyboard friendly form for making reservations

Jokinen:
Varaamo currently uses FullCalendar as its UI for selecting a time to be reserved. FullCalendar is not very keyboard friendly, and in turn does not offer support for things such as screen readers either. In this PR, I add a basic form with three inputs, which allow tab users to select a date and time to make a reservation for.

This form is situated before the FullCalendar component in the resource page. It becomes visible when it has `:focus-within`--this way it doesn't distract users who don't use tab. Seeing keyboard users may find this behaviour unexpected.

The form has three inputs: `date`, `start` and `end` time.

* The `date` input relies on the `type="date"` input the browser provides. The date pickers that Varaamo uses are not reliably accessible based on audits. The `type="date"` picker is hopefully accessible enough on major browsers. When the user changes the date, time selections are reset.
* The `start` and `end` inputs are native selects. I attempted to use HDS's `Dropdown` but ultimately was not able to make it work for this usecase. `start` and `end` inputs' options are disabled based on each other's values, and some other conditions. This way, I hope, that this approach to selecting a time is at least remotely convenient to use. By disabling times that are not OK, we hopefully limit the amount of error states that can be reached while also making the choice easier for the user. However, I am concerned whether this approach is clear enough about why options are disabled.
  * `start` option is disabled when
    * it's in the past
    * it's after selected `end time`
    * it would result in a reservation that break `minPeriod` or `maxPeriod`
    * there's a reservation between the `start option` and the selected `end time`
    * all start options are enabled when both a starting and ending time are selected--when a user selects a start option, the selected end time is reset
  * `end` option is disabled when
    * it's in the past
    * it's before selected `start time`
    * it would result in a reservation that break `minPeriod` or `maxPeriod`
    * there's a reservation between the `end option` and the selected `start time`
    * all start options are enabled when both a starting and ending time are selected--when a user selects an end option, the selected start time is reset


src/common/calendar/TimePickerCalendar.js
```diff
@@ -396,7 +396,7 @@ class TimePickerCalendar extends Component {
               click: () => onDateChange(this.calendarRef.current.calendar.view.initialNowDate),
             },
           }}
-          datesRender={this.onDatesRender}
+          // datesRender={this.onDatesRender}
```
badrange, line 399: Any special reason for leaving this line commented out instead of deleted? In that case a comment to explain why it is still her is helpful!

src/domain/resource/resourceKeyboardReservation/ResourceKeyboardSlotPicker.js
```diff
@@ -0,0 +1,311 @@
...
+function getClassName(className) {
+  return `${rootClass}__${className}`;
+}
+
+const classes = {
+  root: rootClass,
+  fieldGrid: getClassName('field-grid'),
+};
+
+function getTime(dateTime) {
```
badrange, line 18: This project uses moment so you could use moment().format()?

badrange: That will also handle summer-winter time and timezone issues for you!

src/domain/resource/resourceKeyboardReservation/ResourceKeyboardSlotPicker.js
```diff
@@ -0,0 +1,311 @@
...
+      ...slot,
+      isResultsInNegativeTime: false,
+      isResultsInOverlappingReservation: false,
+      isResultsInUnderMinTime: false,
+      isResultsInOverMaxTime: false,
+    }));
+  }
+
+  // eslint-disable-next-line max-len
+  const getIsResultsInNegativeTime = slotStart => new Date(slotStart).getTime() > new Date(selectedBounds.end).getTime();
```
badrange, line 57: Will this cause time zone issues if the computer is set to Stockholm time? Could be better to use moment.tz?

src/domain/resource/resourceReservationButton/ResourceReservationButton.js
```diff
@@ -0,0 +1,134 @@
+import * as React from 'react';
+import PropTypes from 'prop-types';
+import Button from 'react-bootstrap/lib/Button';
+import get from 'lodash/get';
+import moment from 'moment';
```
badrange, line 5: Isn't it so that you should use moment.tz ?

badrange: Looking at the code in Varaamo I see that I'm probably confused. Good to verify that no bugs arise if you set your computer to a time in another timezone

src/common/calendar/TimePickerCalendar.js
Jokinen: I actually just forgot to uncomment after debugging. I had a problem with an infinite loop. I fixed a similar issue for `resourcemap` so I attempted to see if that was the root of my issue and forgot this in. I don't think this prop does anything, but it's not within the scope of these changes to tamper with it so I'll just remove the comment for now.

src/domain/resource/resourceKeyboardReservation/ResourceKeyboardSlotPicker.js
Jokinen: I'm not sure what issue you are referencing here. I tested it out with a Mumbai and Sao Paulo timezone to verify behaviour.

Generally speaking, I don't think that there should be huge danger for things going wrong because of timezones. JS dates try really hard to be timezone aware, in that sense it comes out of the box. From general code health, I think it's OK.

But there is the more usability oriented question which has to do with the behaviour. Currently times are shown in the locale time of the browser, which may actually be inoptimal. If you were referencing this, then you I agree with you. We could use `moment.tz` here to force the timezone to Helsinki time, which would allow visitors from abroad to make reservations in locale time, which maybe makes more sense. May be a bit overkill as well, I'll have to look into what it entails.

src/domain/resource/resourceKeyboardReservation/ResourceKeyboardSlotPicker.js
Jokinen: I try to actively avoid using `moment` to make it easier to refactor out :D But you are right that here it's a borderline case--I wouldn't go through much more effort than this and not always even this much. I used moment on the parent component to format dates for instance.

Jokinen:
I replaced the implementation of `getTime` with one that uses moment, and hardcodes the time zone into the app timezone (Helsinki). This way users will see times rendered in Helsinki time regardless of the time zone they use.

PR #1156, opened 2020-06-10
VAR-391 | Fix buttons being enabled when they should not

Jokinen:
These can only be tested in production.

On the user reservations page, some buttons would be enabled when they should not be. The particular case is this:
- user is a regular user (i.e. not admin or staff)
- the reservation the buttons are for is "paid" (in other words, it has products)

There was a two-fold bug in this behaviour:
1) The resource object was not properly accessible to modals, because it was not saved into redux state
2) The resource object that user reservations page could provide was the "simplified resource object" that's returned in the fields of `reservation` objects. This resource object variation omits most fields, including the products field, which is used for checking whether the reservation is paid or not


app/pages/user-reservations/UserReservationsPage.js
```diff
@@ -186,9 +202,55 @@ class UnconnectedUserReservationsPage extends Component {
...
+              resource,
+            };
+          },
+        );
+
+        // Here we are duplicating the loaded reservations into the
+        // redux state. This makes them discoverable for some action
+        // buttons and modals which are heavily tied to the redux state.
+        // This is a relatively brittle approach to handle this and can
+        // likely cause new issues later on.
```
badrange, line 244: I appreciate your honesty!

PR #1159, opened 2020-06-11
Take codeformatter into use

Jokinen:
This commit adds prettier to the project. Previously this project did
not have a code formatter which caused some unnecessary whitespace and
other such changes.


Jokinen:
I've not merged this for now, as the dev team is changing.

PR #1169, opened 2020-06-16
Fix number of participants being mandatory

Jokinen:


app/pages/reservation/reservation-information/ReservationInformationForm.js
```diff
@@ -35,7 +35,7 @@ const validators = {
   numberOfParticipants: (t, { numberOfParticipants }) => {
     const isNumber = !isNaN(numberOfParticipants);
 
-    if (!isNumber) {
+    if (numberOfParticipants && numberOfParticipants && !isNumber) {
```
tuovinensanttu, line 38:
```suggestion
    if (numberOfParticipants && !isNumber) {
```
Little type here.

PR #1170, opened 2020-06-16
VAR-380 | Add required checkbox for specific terms

Jokinen:


app/pages/reservation/reservation-information/ReservationInformationForm.js
```diff
@@ -64,14 +64,23 @@ const maxLengths = {
...
-    : 'ReservationForm.termsAndConditionsError';
+  const errorLabels = {
+    paymentTermsAndConditions: 'ReservationForm.paymentTermsAndConditionsError',
+    termsAndConditions: 'ReservationForm.termsAndConditionsError',
+    specificTerms: 'ReservationForm.specificTermsError',
+  };
+
+  if (!errorLabels) {
+    return errorLabels.termsAndConditions;
+  }
```
tuovinensanttu, line 81: Isn't `errorLabels` going to be true every time? Should this check `field` instead?

Jokinen: Yeah, this was a mistake. I wanted to retain the behaviour of the original function which would return the `termsAndConditions` error in case checks did not match. So I wanted to to check whether the error label existed, and if not, return the `termsAndConditions` error instead. I've added a fix commit.

PR #1171, opened 2020-06-17
VAR-359 | Add contrast controls

Jokinen:
Screenshot 2020-06-17 at 9 37 05
Screenshot 2020-06-17 at 9 36 43


Jokinen:
@badrange I've talked with our PO about it, and it's something we want to do. For instance https://palvelukartta.hel.fi/en/ has added this feature as well.

Jokinen:
@tuovinensanttu That's a good point. I've tuned up the contrast to a degree where the edges of fonts may get "crispy". It's a weakness in the approach I've used here. I can't quickly come up with ways to avoid it.

badrange:
> @badrange I've talked with our PO about it, and it's something we want to do. For instance https://palvelukartta.hel.fi/en/ has added this feature as well.

👍 

I hope it will be useful!

PR #1183, opened 2020-08-18
Fix DateRangePicker issues

ahie:
- Fixes translations displayed for overlapping and unavailable reservations
- Adds missing styles for overnight reservations
- Makes date range validators inclusive


Nizar-Rahme (approved):
LGTM!

PR #1186, opened 2020-08-31
VAR-431 | Accessibility profile menu

Nizar-Rahme:
## Description :sparkles:
- Implemented accessibility settings menu
- Fetch accessibility options from an endpoint
- Store the selected values in localStorage

### Closes :no_good_woman:
[VAR-431](https://helsinkisolutionoffice.atlassian.net/browse/VAR-431): Implement the accessibility menu component

### Automated tests :gear:️
https://github.com/City-of-Helsinki/varaamo/blob/45aa425b58761a4abc37f7e492363870dd5b9aa9/app/utils/__tests__/localStorageUtils.test.js#L3

## Screenshots :camera_flash:
image


src/domain/header/MainNavbar.js
```diff
@@ -98,6 +101,26 @@ class MainNavbar extends React.Component {
...
+          <Nav pullRight>
+            <TabbableNavDropdown
+              as="li"
+              id="user-nav-dropdown"
+              renderToggle={(props) => {
+                return (
+                  <a {...props} className="app-MainNavbar__selection" href="#" type="button">
+                    {t('Accessibility settings')}
+                    <span className="app-MainNavbar__selection-overview">
+                      {`${(getFromLocalStorage(ACCESSIBILITY_LOCAL_STORAGE_KEY) || []).length} ${t('Selections')}`}
```
ahie, line 113: > getFromLocalStorage(ACCESSIBILITY_LOCAL_STORAGE_KEY) || []

Could abstract this to its own utility function? Same for

> storeInLocalStorage(ACCESSIBILITY_LOCAL_STORAGE_KEY, selectedViewpoints)

in another component.

src/domain/header/MainNavbarContainer.js
```diff
@@ -1,10 +1,32 @@
...
+const apiClient = new ApiClient(settings.ACCESSIBILITY_API_URL, false);
+
+const MainNavbarContainer = (props) => {
+  const [viewpoints, setViewpoints] = useState([]);
+
+  useEffect(() => {
+    apiClient.get('viewpoints').then(
+      ({ error, data }) => {
+        if (error || viewpoints.length) return;
+        if (Array.isArray(data)) setViewpoints(data);
```
ahie, line 21: Should we blacklist the first (id 0?) viewpoint here, which is "Choose accessibility perspective"?

src/domain/header/AccessibilityMenu.js
```diff
@@ -0,0 +1,92 @@
...
+  const handleSubmit = (e) => {
+    e.preventDefault();
+
+    storeInLocalStorage(ACCESSIBILITY_LOCAL_STORAGE_KEY, selectedViewpoints);
+    handleClose();
+  };
+
+  return (
+    <form className="app-AccessibilityMenu" onSubmit={handleSubmit}>
+      <div className="app-AccessibilityMenu__btn-wrapper">
```
ahie, line 39: > <div className="app-AccessibilityMenu__btn-wrapper">

This suggestion diverges from the designs a bit, but would it make sense to make this top wrapper sticky and remove the bottom one? Right now both submit buttons have the same functionality, so it's a bit confusing to have one called "save settings" and another that just has the text "save".

I think the miro designs specify that the menu doesn't automatically close on save, so this top wrapper would contain the text "changes saved" after click, and possibly reset once the selection has changed, but IMO that's also a bit questionable as closing the menu would then require you to explicitly click "Close" every single time.

ahie (changes requested):

src/domain/header/MainNavbar.js
Nizar-Rahme: Done

src/domain/header/MainNavbarContainer.js
Nizar-Rahme: Done

src/domain/header/AccessibilityMenu.js
Nizar-Rahme: Yes, I agree with you. I removed the bottom buttons group.
I am not also sure whether to make the "Save" button closes the menu or not, for now, I left as it is 🤷 .

src/domain/header/utils.js
```diff
@@ -0,0 +1,7 @@
+import { getFromLocalStorage, storeInLocalStorage } from '../../../app/utils/localStorageUtils';
```
ahie, line 1: Optional: could move these to `common/utils`, as most of the utility functions live there.

src/domain/header/AccessibilityMenu.js
ahie: Seems fine to me :).

src/domain/header/_accessibilityMenu.scss
```diff
@@ -0,0 +1,33 @@
+.app-AccessibilityMenu {
+    padding: 15px 30px;
+    width: 420px;
+    overflow: scroll;
```
ahie, line 4:
```suggestion
    overflow: hidden;
```

Otherwise the menu will have two scrollbars.

ahie (changes requested):

src/domain/header/utils.js
Nizar-Rahme: Do you mean something like `/src/common/localStorage/utils.js`?

src/domain/header/_accessibilityMenu.scss
Nizar-Rahme: Good catch!

src/domain/header/utils.js
ahie: `/app/utils/accessibilityUtils.js` was thinking of initially.

Nizar-Rahme: Ok, I will move them

Nizar-Rahme: I was just thinking @ahie these two util functions are too abstract to be tied to only accessibility functionality, what do you think?
They just store and get values from local storage.

ahie (approved):
LGTM 👍

PR #1188, opened 2020-09-04
VAR-437 | Centralize z-index values in one sass partial

Nizar-Rahme:
## Description :sparkles:
Gather all `z-index` values in one Sass partial

### Closes :no_good_woman:
**[VAR-437](https://helsinkisolutionoffice.atlassian.net/browse/VAR-437)**: Centralise z-index values in one sass partial


app/assets/styles/_z-index.scss
```diff
@@ -0,0 +1,56 @@
+$z-index: (
```
ahie, line 1: Optional: could clean the values here, but keeping their order the same.

ahie (approved):
LGTM! Just left an optional comment.

app/assets/styles/_z-index.scss
Nizar-Rahme: Good suggestion 👍 
Done!

PR #1189, opened 2020-09-09
VAR-438 | Use HDS spacing values

Nizar-Rahme:
## Description :sparkles:
Implement one SASS partial for defining all spacing values and use only these values across the application.
### Closes :no_good_woman:
[**VAR-438**](https://helsinkisolutionoffice.atlassian.net/browse/VAR-438): Centralise padding and margin values in one sass partial

### Manual testing :construction_worker_man:
Open the application and navigate through all the pages, you can compare the interface with the production instance.

### DIFF
--- a/app/assets/styles/_application-variables.scss
+++ b/app/assets/styles/_application-variables.scss
@@ -66,7 +66,7 @@ $button-horizontal-padding:    15px;
 
 $app-ResourceCard-height: 180px;
 $app-SearchPage-max-width: 800px;
-$app-SearchPage-margin: 10px;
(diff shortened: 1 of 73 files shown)


ahie (approved):
Went through at least a majority of different views and everything seems to work fine 👍 .

Also tried looking for any remaining places by grepping with `(margin|padding|height|width)(-[a-z]*)?:.*(px|rem|em|ex|ch|vv|vh|%|vmin|vmax|cm|mm|in|pt|pc)` and couldn't find any.

PR #1191, opened 2020-09-25
VAR-429, VAR-432 | Implement sorting resources by accessibility

ahie:
## Description :sparkles:

Adds sorting by accessibility to the resource search page.

This commit additionally connects the selected accessibility preferences
to the redux controlled state. This enabled to e.g. update the
resource listing's accessibility icons instantly as preferences are
changed.

### Closes :no_good_woman:

[VAR-429](https://helsinkisolutionoffice.atlassian.net/browse/VAR-429)
[VAR-432](https://helsinkisolutionoffice.atlassian.net/browse/VAR-432)

### Automated tests :gear:️

`app/state/reducers/__tests__/accessibilityReducer.test.js `

### Manual testing :construction_worker_man:

- On the front page, search for resources.
- Once the results are listed, toggle sorting by accessibility from the dropdown menu.
- The resources should be refetched, with resources that have shortcomings for your accessibility preferences listed last.


src/domain/resource/card/ResourceCard.js
```diff
@@ -187,9 +189,14 @@ class ResourceCard extends React.Component {
...
-                  nShortcomings: accessibilityShortcomingsCount,
-                })
+                ? t(
+                  accessibilityShortcomingsCount === 1
+                    ? 'ResourceInfo.foundAccessibilityShortcoming'
+                    : 'ResourceInfo.foundAccessibilityShortcomings',
+                  {
+                    nShortcomings: accessibilityShortcomingsCount,
+                  },
+                )
```
Nizar-Rahme, line 199: Wondering whether `react-intl` has an official way to handle pluralization.

app/state/reducers/ui/accessibilityReducer.js
```diff
@@ -17,6 +18,12 @@ function toggleContrast(state) {
   return { ...state, isHighContrast: !state.isHighContrast };
 }
 
+function setAccessibilityPreferences(state, preferences) {
+  const newPreferences = Array.isArray(preferences) ? preferences : [];
+  setSelA11yPref(newPreferences);
```
Nizar-Rahme, line 23: I think should be pure functions and we should avoid performing side effects inside them [according to Redux docs](https://redux.js.org/basics/reducers#handling-actions).
Probably the subscriber should handle persisting and retrieving the state from the `localStorage`(https://stackoverflow.com/a/35675304/9441362)

Nizar-Rahme (changes requested):

src/domain/resource/card/ResourceCard.js
ahie: There are components, but I don't think there's a simpler way with the `t` function. In any case you need to provide the translations manually for singular and plural forms.

app/state/reducers/ui/accessibilityReducer.js
ahie: This is true, moved it to a separate middleware.

Nizar-Rahme (approved):
LGTM 👍

PR #1192, opened 2020-09-29
VAR-444 | Add cancel reservation reason to "Reservation Information" modal

Nizar-Rahme:
## Description :sparkles:
Add cancel reason type, category, and explanation to Reservation Information modal

## Screenshots :camera_flash:
image


ahie (approved):
Could add a test for the array(/multiple value) support, otherwise LGTM. The optional test could be added in a separate PR.

Nizar-Rahme:
Good point! I will add UT in a follow-up pull request.

PR #1241, opened 2022-06-03
TILA-1687 dependency updates

kieferbonk:
- Replace `lightbox-react` with `react-image-lightbox`
  - Replace `node-sass` with `dart-sass`
  - Update deps


vikoivun (approved):
Looks clean. As a bonus it would be nice document why the updates were done (security issue, functionality broken).

